PR #8, opened 2015-01-11
enable ES6 support and make it work with gulp 4.0.0-alpha

geekflyer:
was discussed in https://github.com/gulpjs/gulp/issues/830.

In this PR I did 2 things:

* merged the gulp-cli master into 4.0, where master was 2 ahead of 4.0, because of ES6 support
* renamed the versioned folder from 4.0.0 to 4.0.0-alpha so that the gulp bootstrap also accepts a local pre-release version of gulp (i.e. gulp 4.0.0-alpha)


phated:
Thanks.  I am going to rebase 4.0 onto master instead of ending up with the merge and duplicate commits.  I also added tests for prereleases in phated/semver-greatest-satisfied-range to make sure it supported alpha, etc.

phated:
Updated the 4.0 branch and also fixed the version number in gulp (it wasn't a well formed semver prerelease).  Thanks for catching these.

PR #11, opened 2015-01-15
bump liftoff and v8flags

tkellen:
this should solve all of the wack installation issues we were
seeing on windows machines, and work correctly with io.js.

- v8flags is now async
- v8flags caches flags at first run, not on installation
- liftoff `nodeFlags` was renamed to `v8flags` for accuracy

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,7 +22,7 @@ var cli = new Liftoff({
   name: 'gulp',
   completions: completion,
   extensions: interpret.jsVariants,
-  nodeFlags: v8flags.fetch()
+  v8flags: v8flags
 });
 
 // exit with 0 or 1
--- a/package.json
+++ b/package.json
@@ -29,12 +29,12 @@
     "chalk": "^0.5.0",
     "gulp-util": "^3.0.0",
     "interpret": "^0.4.1",
-    "liftoff": "^1.0.0",
+    "liftoff": "^2.0.0",
     "minimist": "^1.1.0",
     "pretty-hrtime": "^0.2.0",
     "semver": "^4.1.0",
     "tildify": "^1.0.0",
-    "v8flags": "^1.0.1"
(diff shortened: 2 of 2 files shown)


contra:
Failed ^

phated:
We are still waiting for some changes.  That being said, I have no idea why appveyor started failing again.  Maybe @heikki can take a look?

heikki:
Install fails because [6to5](https://github.com/6to5/6to5/blob/master/package.json#L65-L67) has [kexec](https://github.com/jprichardson/node-kexec) as optionalDependency and that doesn't support windows. But..

> If a dependency can be used, but you would like npm to proceed if it cannot be found or fails to install, then you may put it in the optionalDependencies object.

Could it be that npm is too old or something?

ps. I'm no windows expert, just making sure I don't have to use ant etc. ever in mixed os projects :smile:

phated:
@heikki installation isn't failing on appveyor, it seems to be a stdout problem again.

heikki:
Ah, sorry. Didn't read carefully enough.

heikki:
It's not only this PR. Current master fails too and even the commit where I "fixed" things originally. Either some sub-sub-dependency has changed something or I'm going crazy.

phated:
holding off on this until @tkellen bumps again.

tkellen:
I'm not in front of my laptop right now--feel free to push a new release,
or I will when I get home (2-3 hours).

phated:
@tkellen I don't have access to `v8flags`.  I'm fine waiting awhile, no hurry.

tkellen:
`2.0.2` released.

phated:
Woot, merging and bumping these

heikki:
I didn't find any reason why windows tests started failing. What's more bizarre is that now they seem to pass again.

PR #14, opened 2015-02-10
Added preferGlobal

callumacrae:
Seems like it fits on this repo. https://docs.npmjs.com/files/package.json#preferglobal

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
   "bin": {
     "gulp": "bin/gulp.js"
   },
+  "preferGlobal": true,
   "man": "gulp.1",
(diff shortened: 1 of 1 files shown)


phated:
I think this will cause the error to show up when it is a dependency of gulp.

callumacrae:
Fair enough. Is there any way to test that without pushing it up to npm?

phated:
you can use npm to install from github with `npm install repo_name/project_name`

phated:
I don't think it is worth it to add preferGlobal here.  The npm warnings are annoying.

PR #20, opened 2015-03-18
fix #19 Support tasks description in master

### LINKED ISSUE #19: Support a `description` property on task functions
If we see a property named `description` on a task function, we should print it when using the `--tasks` flag.

ref gulpjs/gulp#591

louisremi:
As requested in [#19](https://github.com/gulpjs/gulp-cli/issues/19)
I also have improved unit tests for --tasks.
If this PR is approved I'm ready to document the feature in gulp docs.

I already love this feature and can't wait for it to land : )

### DIFF
--- a/index.js
+++ b/index.js
@@ -137,6 +137,7 @@ function handleArguments(env) {
 function logTasks(env, localGulp) {
   var tree = taskTree(localGulp.tasks);
   var padding = 0;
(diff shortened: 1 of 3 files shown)


louisremi:
Here's how it looks, by the way

gulp tasks tree with description

phated:
Thanks! Published as 0.2.0

PR #21, opened 2015-03-18
Fix #19 support tasks description in 4.0

### LINKED ISSUE #19: Support a `description` property on task functions

louisremi:
In addition to [#20](https://github.com/gulpjs/gulp-cli/pull/20), support task description in 4.0
Complete with unit tests


phated:
I had to cherry-pick this into 4.0 due to rebasing.  Thank you so much for the PR!!!

PR #23, opened 2015-04-10
Fix badges

austinpray:


contra:
Thanks!

austinpray:
@contra having the same problem as you guys with the code coverage and liftoff. I'll let you know/PR if I figure it out.

PR #28, opened 2015-06-20
[4.0] Fixes gulp --tasks

clintwood:
This PR fixes gulpjs/gulp#1112

Caveat: Not sure what output is desired, so this is a best guess which works for now.


clintwood:
Sorry don't run jslint (I use eslint) but seemed to have resolved linting errors I introduced, it's now failing on files I didn't touch...

phated:
you approached this in a very roundabout way.  you can just use `gulpInst.task(task)` because task is a getter and a setter.

clintwood:
@phated, Ahhh, that's what I was looking for :D
Thanks will update PR!

PR #32, opened 2015-09-10
4.0 introduce gulplog

kahlil:
* removed `gulp-util` dependency
* added `gulplog` dependency
* use `gulplog` instead of `gutil.log`
* wire up the `gulplog` event listeners
* filter `gulplog` event listeners according to log level
* added `.editorconfig` file

### DIFF
--- a/lib/shared/cliOptions.js
+++ b/lib/shared/cliOptions.js
@@ -71,6 +71,7 @@ module.exports = {
       'even when color support is detected.'),
   },
   silent: {
+    alias: 's',
     type: 'boolean',
     desc: chalk.gray(
       'Suppress all gulp logging.'),
@@ -79,5 +80,13 @@ module.exports = {
     type: 'boolean',
     desc: chalk.gray(
       'Continue execution of tasks upon failure.'),
+  },
+  loglevel: {
+    alias: 'l',
+    type: 'boolean',
+    count: true,
+    desc: chalk.gray(
+      'Set the loglevel. -l for least verbose and -llll for most verbose.' +
+      ' -lll is default.'),
   },
 };
--- a/lib/shared/log/toConsole.js
+++ b/lib/shared/log/toConsole.js
@@ -22,7 +22,7 @@ function toConsole(log, opts) {
     levelsToListen = levels.slice(levels.indexOf(opts.logLevel));
   }
 
-  levels.forEach(function(level) {
+  levelsToListen.forEach(function(level) {
     log.on(level, fancyLog);
   });
 }
--- a/lib/shared/log/toConsole.js
+++ b/lib/shared/log/toConsole.js
@@ -1,0 +1,25 @@
+'use strict';
+
+var chalk = require('chalk');
+var dateformat = require('dateformat');
+
+var levels = [
+  'info',
+  'error',
+  'warn',
+  'debug',
+];
+
+function shellOut() {
+  var time = '[' + chalk.grey(dateformat(new Date(), 'HH:MM:ss')) + ']';
+  process.stdout.write(time + ' ');
+  console.log.apply(console, arguments);
+}
+
+function toConsole(log) {
+  levels.forEach(function(level) {
+    log.on(level, shellOut);
+  });
+}
+
+module.exports = toConsole;
--- a/lib/shared/log/toConsole.js
+++ b/lib/shared/log/toConsole.js
@@ -1,7 +1,6 @@
 'use strict';
 
-var chalk = require('chalk');
-var dateformat = require('dateformat');
+var fancyLog = require('fancy-log');
 
 var levels = [
   'info',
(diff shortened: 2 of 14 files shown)


kahlil:
@phated @contra  plz review

kahlil:
Not sure where you are at with the failing tests. They are failing in 4.0 branch as well. I guess I should add some tests for the logging as well?

kahlil:
haha i just realized I was wrong, I destroyed the tests. will fix. sorry bout that. also will add testing for logging.

package.json
```diff
@@ -42,6 +43,7 @@
   "devDependencies": {
     "code": "^1.2.1",
     "coveralls": "^2.7.0",
+    "fancy-log": "^1.0.0",
```
phated: this should be a regular dependency

contra:
Reminder to squash all of these after reviewing - thanks again!

lib/shared/log/toConsole.js
```diff
@@ -0,0 +1,30 @@
...
+
+var fancyLog = require('fancy-log');
+
+// The sorting of the levels is
+// significant.
+var levels = [
+  'debug', // Logs all levels.
+  'info',  // Logs info, warn and error.
+  'warn',  // Logs warn and error.
+  'error', // Logs error.
```
phated: after reviewing a bunch of the old threads about "a real logger" it seems that @contra wanted errors to be output on stderr.  I think I will add another method on `fancy-log` that calls `console.error` instead of `console.log`

phated:
@contra any opinion on the CLI syntax for adjusting log level?

lib/shared/log/toConsole.js
kahlil: @phated in this case, we should refactor `fancy-log` to handle errors with `console.error` right?

package.json
kahlil: @phated thanks! fixed.

kahlil:
@contra didn't know about the squashing, thanks. And like @phated said: any input on the log-level syntax? 😀

phated:
I am still leaning towards `-vvvv` for most verbose using a yargs "count" variable that defaults to 3 (info level) and goes up to 4:
* count === 1 is error level
* count === 2 is warn level
* count === 3 is info level
* count === 4 is debug level

We can also keep the `--silent` option to disable all logging

tkellen:
:+1: to -vvvv style log level handling!

lib/shared/log/toConsole.js
phated: @kahlil yep, I've updated `fancy-log`. The `log.error()` API is available in 1.1.0 - https://github.com/phated/fancy-log

kahlil:
@phated @tkellen ok then I'll just go ahead and implement that.

phated:
@kahlil awesome, thanks!  I think I've unblocked everything to update this PR.  This is great work. Thanks again!

kahlil:
Thanks @phated, glad to help!

I just realized we have a little problem though: the alias `-v` is reserved for the `--version` option already. 
Any thoughts on how to go about that? Should we use `-l` instead?

kahlil:
@phated implemented it with `-l` to `-llll` for now. Let me know how you want to deal with the conflict for the `-v` alias and what you want to have changed.

lib/shared/log/toConsole.js
```diff
@@ -0,0 +1,39 @@
...
+  // the default count for opts.loglevel to 2
+  // instead of 0 when we configure count: true
+  // in the cliOptions.
+  if (opts.loglevel > 2) {
+    loglevel = opts.loglevel - 2;
+  }
+
+  levelsToListen = levels.slice(0, loglevel);
+
+  levelsToListen.forEach(function(level) {
```
phated: unfortunately, we can't just do this anymore because fancy-log has the `fancyLog.error` method to log to stderr.  I suppose the forEach function body could check if level === 'error'

kahlil:
Got it. Will fix.

lib/shared/log/toConsole.js
kahlil: done

kahlil: ah not done, have to fix the test ;)

kahlil: @phated ok now it's done :smile:

lib/shared/cliOptions.js
```diff
@@ -80,4 +81,12 @@ module.exports = {
     desc: chalk.gray(
       'Continue execution of tasks upon failure.'),
   },
+  loglevel: {
+    alias: 'l',
```
phated: capital `-L` and capital `-S`?  @tkellen @contra thoughts?  Could change `loglevel` to `verbose` and use capital `-V` if you don't like those...

phated:
Seems to have missed the 4.0.0 folder.  I'm tackling that now

phated:
Updated.  Thanks again for the PR!

kahlil:
You're welcome :smile: thanks for taking care of the yargs update!

Issue #35, opened 2015-09-19
Support recursive tree output for --tasks


sandipchitale:
Something like:
```javascript
        function printTaskTree(tasks, taskKey, indent) {
		    var taskObject;
		    taskObject = tasks[taskKey];
		    if (taskObject) {
		        console.log(indent + (taskObject.name ? taskObject.name : taskKey));
		        if (taskObject.dep) {
		            taskObject.dep.forEach(function(name) {
		                printTaskTree(tasks, name, indent + '  ');
		            });
		        }
		    }
		}

        printTaskTree(gulp.tasks, 'default', '');
```

contra:
Send a PR?

phated:
This can be done in userland with the exposed `gulp.tree()` method.  Not likely to add this to the CLI

phated:
As per https://github.com/gulpjs/gulp-cli/pull/36#issuecomment-142103723, undertaker builds the recursive tree, it shouldn't be the responsibility of the `taskTree` method to do that.

phated:
The test has been updated to be less confusing in https://github.com/gulpjs/gulp-cli/commit/1c0da0d35a81ac1a0e0e4e3296026bd06cd43d12

PR #36, opened 2015-09-20
Reduce full tree

sandipchitale:
Pull request for implementation of:

https://github.com/gulpjs/gulp-cli/issues/35

### DIFF
--- a/lib/taskTree.js
+++ b/lib/taskTree.js
@@ -12,7 +12,7 @@ module.exports = function(tasks) {
     prev.nodes.push(node);
     return prev;
   }
-  
+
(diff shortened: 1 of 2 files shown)


phated:
It seems that the test for this has confused you.  The `treeTree` method just outputs a given task tree, it doesn't need to build the dependency tree.  The dependency tree is built and stored inside `undertaker` and does this recursive building already.  I will adjust this test so it is less confusing.

phated:
The test has been updated in https://github.com/gulpjs/gulp-cli/commit/1c0da0d35a81ac1a0e0e4e3296026bd06cd43d12

PR #47, opened 2015-12-18
register any function that is exported from a gulpfile

phated:
@thejameskyle As per gulpjs/gulp/issues/1305 - can you try this out?

@contra you good with this functionality being added?


thejameskyle:
Tried getting this working, and wasn't able to https://github.com/thejameskyle/__gulp-exports

Not sure if I'm missing something. I'm depending on this branch: https://github.com/thejameskyle/__gulp-exports/blob/master/package.json#L16

phated:
@thejameskyle I didn't add it for gulp 3.9 but maybe I could/should

thejameskyle:
It doesn't work for 4.0 either: https://github.com/thejameskyle/__gulp-exports/tree/4.0

phated:
@thejameskyle yeah, digging into it...npm doesn't seem to know how to dedupe github dependencies, so you end up with the gulp-cli#4.0 branch as a dependency of gulp, which is used instead of the gulp-cli you install as a dep

phated:
@thejameskyle deleting `node_modules/gulp/node_modules/gulp-cli` allows it to work

thejameskyle:
Ah yup, manually running `npm dedupe` works too. Thanks! This is awesome

contra:
Yeah this makes perfect sense

phated:
Was easy enough to backport too!  3.7 and up will support this if someone is using this CLI separately

PR #50, opened 2015-12-24
Modify for gulp --tasks

sttk:
I tried to achieve @tylersticka's  second image in gulpjs/gulp#1294 and finished.

  * Don't hide series/parallel tasks.
  * Can change task tree depth by '--depth' option.
  * Can set a task description with .description property, as follows:   
    ```
       gulp.task('task-name', fn).description = 'task-description';
    ```  
    or  
    ```
       gulp.task('task-name', fn).description = { '': 'task-description', '--option1': 'task-option-1', ...  };
    ```


tkellen:
This looks awesome @sttk!

sttk:
Sorry for some errors but I fixed them all.

phated:
Instead of adding flags into the `description` property, I think we should use a `flag` property that has `flag: description` key-value pairs.  That avoids the empty string key for the task description which looks really clunky

sttk:
> I think we should use a flag property that has flag: description key-value pairs.

Is it right this example?

```
function build(cb) { cb() }
build.description = 'Build all the things!';
build.flag = {
  '--dev': 'un-minified',
  '--production': 'compressed into single bundle'
};
gulp.task(build);
```

phated:
Yep!

sttk:
I've finished modifying the program!

sttk:
I've remade `logTasks` function in gulp-cli, because I brought the changes in undertaker here and made undertaker no change.

Modification from master branch:

  * Enable to display a descriptions and flags (options) for each task. they are fetched from task function's properties.

  * Top level tasks are listed in alphabetical order.

  * Enable to control the depth of the task tree. If the depth is null or not number, the full tree is displayed.

  * The API of `logTasks` was changed from `(tree, getDescription)` to `(tree, depth, getTask)`.

phated:
Awesome!  I hope to review this soon.  Thanks

phated:
@sttk I looked over this and everything looks solid.  However, you need to update the ^3.7 and 4.0.0-alpha.1 files also because they use the same shared `logTasks` method.

sttk:
@phated Thanks. And sorry that I forgot those versions.
I've added the modifications to those versions as well. 

Modification:

* ^3.7.0

    * The API of `logTasks` was changed.
    * `taskTree` function was changed to adapt the structure for `logTasks`.

* ^4.0.0-alpha.1

    * The API of `logTasks` was changed.

erikkemperman:
@sttk @phated  I ran across that other [issue](https://github.com/gulpjs/gulp/issues/1294) and was just trying out this patch. Looks really sweet!

One minor thing though, but maybe this is just because I tend to have slightly transparent terminals: the grey color of the timestamps and the tree branches is maybe a little bit too dark?

And actually I would love for the timestamps to be suppressable altogether... That might allow for some cleaner tests as well (which now can't simply compare with a fixture because the timstamps will be different each run). But I suppose that is down to `fancy-log`?

sttk:
About the timestamps I displayed it according to the second image in #1294, but I also think better to remove it because of more display width.

May I remove the timestamps?

sttk:
About the color problem I have an idea that is to add a color table which is user customizable, and this table can be used for other logs. (Though I'd like to modify this on another PR.)

For example:
```
(gulp-cli/lib/shared/log/colorTable.js)
var chalk = require('chalk');
modules.exports = {
  ...
  tasklist: {
    simple: { ... },
    tree: { branch: chalk.gray, task:{ name: chalk.cyan, flag: chalk.magenta, ...
  },
  ...
}
---
(gulp-cli/lib/versioned/^4.0.0-alpha.2/index.js)
...
var colorTable = require('../../shared/log/colorTable');
   ...
   gulpInst = require(env.modulePath);
   gulpInst.colors = colorTable;
   ...
--
(gulp-cli/lib/shared/log/tasks.js)
...
function logTasks(tasks, depth, colorTable, getTask) {
  ...
  var line = 
    colorTable.tasklist.tree.branch(branch) +
    colorTable.tasklist.tree.task.name(taskname) +
    spaces +
    colorTable.tasklist.tree.task.description(description);
  console.info(line);
  ...
```

And users can customize these colors as they like, as follows:
```
(In user's custom-colors.js)
var gulp = require('gulp');
var chalk = require('chalk');
gulp.colors.tasklist.tree.branch = chalk.black;
gulp.colors.tasklist.tree.task.name = s => chalk.bold(chalk.red(s));
...
--
(In user's gulpfile.js)
var gulp = require('gulp');
require('./custom-colors');
...
```

erikkemperman:
@sttk For what it's worth, I've looked briefly into suppressing the timestamps -- it would require a change to the `fancy-log` repo, over [here](https://github.com/js-cli/fancy-log).

tylersticka:
Hello! Designer from gulpjs/gulp#1294 here. :wave: 

Just tried this PR out locally, and it looks gorgeous! Awesome job, @sttk !

My only point of nitpicky design feedback: I think the descriptions could use _two_ leading spaces instead of just one.

```
# One space
[00:00:00] ├── foo I am a description
[00:00:00] └── bar I am also a description

# Two spaces
[00:00:00] ├── foo  I am a description
[00:00:00] └── bar  I am also a description
```

I know that's _totally_ nitpicking, but I think that little change might help differentiate the name column from the description column a bit more. (FWIW, my mockup in gulpjs/gulp#1294 uses two spaces.)

With regard to @erikkemperman's feedback, I like @sttk's idea of addressing this via optional user customization rather than in the default output itself. Because Terminal colors are customizable (the gray is perfectly readable for me), we'd be playing whack-a-mole trying to satisfy everyone.

erikkemperman:
@sttk @tylersticka You're absolutely right, I just tweaked my terminal colors to compensate for slightly transparent (black) background and it looks great. Likewise with unadjusted ANSI colors on opaque black.

One thing I didn't notice before is the alphabetical sorting. I had to drop the `undertaker-forward-reference` registry into my gulpfile to be able to have my tasks in a logical (for me) order. Of course that wasn't alphabetical... Would you consider making that optional?

But yeah, nitpicking -- it looks awesome to me!

phated:
I don't like the idea of adding an API to gulp core for colors since it is strictly a CLI problem.  Maybe we should consider adding a `--colors` option that accepts a json/js file with color mappings?

erikkemperman:
@phated Do you have strong feelings about the unconditional timestamps in `fancy-log`?

phated:
@erikkemperman no strong feelings, it was just a hold-over from the old gulp-util stuff.

phated:
@sttk please remove the color commit when you get a chance.  I think we can defer the timestamp stuff until the next version too.  I want to release this soon.

sttk:
@phated The color mapping is just an idea yet. I won't contain it in this PR.
I didn't commit it to the branch for this PR and the above is another branch.

I will address what you all pointed out here in new PR after the release of the next version.
I'm looking forward to the release.

phated:
@tylersticka was it your intention for the `...` to be automatically inserted by us for each flag description?  I assumed it was just a mock of something the end user could do (in the same vein as putting a dash at the start)

tylersticka:
@phated It was my intention. I don't think most people know how to type the horizontal ellipsis character, and I felt the flag descriptions needed something subtle to make them feel like a continuation of the task description.

That said, I kind of expected y'all to push back on the idea. :wink: I understand the design intent might crumble beneath the realities of how developers end up actually _using_ the feature. And ultimately I'm not the one who would have to field or support complaints or requests to override the prefix.

So while that's my intent, I'm flexible about it. You'd know better than I how helpful or crazy-making the ellipsis might be to people using the Gulp CLI.

phated:
@tylersticka thanks for the explanation.  I am fine having it in there for initial release and I'm not sure what the overall feedback will be, but I was just wondering if that was the intention.

phated:
Until chalk releases their new version with 256 and 16m color support, we are stuck with the 8 colors they currently expose.  I think we should keep the branches, child task names and descriptions white to deal with this for the time being.  I really want to come back to this after their release and use the colors @tylersticka used in the mockup.

lib/versioned/^4.0.0/index.js
```diff
@@ -48,8 +48,8 @@ function execute(opts, env) {
       tree = gulpInst.tree({ deep: true });
       tree.label = 'Tasks for ' + chalk.magenta(tildify(env.configPath));
 
-      return logTasks(tree, function(task) {
-        return gulpInst.task(task).description;
+      return logTasks(tree, opts.depth, function(taskname) {
+        return gulpInst.task(String(taskname.toString()));
```
phated: @sttk why is this wrapped in `String()` and having `toString()` called upon it?

tylersticka:
> I really want to come back to this after their release and use the colors @tylersticka used in the mockup.

FWIW, my mockup just used the eight Terminal colors I use on my own machine, which are based on [Seti Syntax](https://atom.io/themes/seti-syntax).

screen shot 2016-01-29 at 12 13 03 pm

phated:
Ah, that makes sense.  I see your magenta is the purple, however your blue is the blue in the picture, whereas we use cyan (blue looks terrible on windows).

lib/versioned/^4.0.0/index.js
phated: I removed these and everything still seems to work.  Just trying to figure out if you encountered something I haven't

tylersticka:
> blue looks terrible on windows

Totally! And I didn't mention it because I think it still looks nice with cyan in my Terminal:

screen shot 2016-01-29 at 12 21 28 pm

lib/versioned/^4.0.0/index.js
sttk: @phated If the `taskname` is a function, gulp (undertaker) regists a new task. There is no such case normally, but I wanted make it sure to avoid the case. So, as you say, there is basically no problem when they removed.

phated:
Other than a few tiny things I did cleanup on, this looks great.  Going to publish as 1.2 shortly

sttk:
@phated Thank you for some appropriate modifications and the merge.

This is the first time I've sended a PR. I'm very glad.
Thanks.

phated:
Just published this as 1.2.0

@sttk amazing work.  Hope you continue to be involved with the gulp community!

PR #52, opened 2015-12-29
Fixed AppVeyor errors.

sttk:
Since some test codes caused errors on AppVeyor, I've fixed them.

Modification:

  * appveyor.yml
    Use `for` statement because the wildcard in `npm test` was not expanded. (But coverage is unavailable).

  * test/completion.js
    Change stdout handling to reading a redirected file because the argument for stdout was empty on node v0.10 for Windows.

  * test/flags-help.js
    Change stdout handling to reading a redirected file because the argument for stdout contained strange strings in the tail on node v0.10 for Windows.

  * test/flags-tasks-json.js
    Add escaping backslashes in `outputString` to eliminate `JSON.parse` errors on Windows, and remove replacing backslashes to slashes in `stdout`.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,7 +21,7 @@ install:
 test_script:
   - node --version
   - npm --version
-  - npm test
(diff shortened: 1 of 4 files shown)


phated:
Thanks!

PR #53, opened 2016-01-02
add displayName support to exported tasks

nmn:
I can't find the exact place with the documentation, but I read somewhere that the task name can be be set using the displayName property on the function. The description property is already working correctly. And displayName is handled correctly elsewhere, (e.g. gulp.parallel)

This small change should fix this.

### DIFF
--- a/lib/shared/registerExports.js
+++ b/lib/shared/registerExports.js
@@ -14,7 +14,7 @@ function registerExports(gulpInst, tasks) {
       return;
     }
 
-    gulpInst.task(taskName, task);
+    gulpInst.task(task.displayName || taskName, task);
   }
 }
 


terinjokes:
You're probably looking for this in the docs: https://github.com/gulpjs/gulp/blob/4.0/docs/API.md#name

nmn:
Yup!

phated:
Can you give me a real world scenario for this?  I'm also including @thejameskyle to see if there is anything I might be missing on this.

thejameskyle:
I'm not aware of any reason you'd want to do this. I can see the case for `displayName` for `gulp.task` tasks if you're doing something dynamically and can't set the `function.name`, but you can't have dynamic ES exports anyways, so I don't see what this helps.

phated:
@thejameskyle can you do something like `css:prod` as the export name?  A lot of people do that for some reason, which is why we support displayName with `gulp.task`

thejameskyle:
Ah, no you cannot. The export name needs to be an identifier just like a function name.

nmn:
The idea is pretty simple. gulp-cli already shows the displayName in many cases. This mostly just makes everything more consistent.

In general, I believe using function.name for the name of the task is a great solution, but as mentioned above often you wanna use hyphens or colons in your task names, and displayName would come in handy in that case.

Also, just to make it clear, this PR has nothing to do with dynamic exports or anything like that. Just slightly more consistent self documentation.

phated:
Backing out this change due to it breaking `export const dist = series(clean, css);` which is a more valid use case than using `displayName` to export things like `css:dist`.  Also going to write some regression tests and publish as 1.2.1

nmn:
tricky...

PR #64, opened 2016-02-28
Avoid locking errors on AppVeyor

sttk:
I encountered locking errors on AppVeyor and the solution is written in following page:

https://www.appveyor.com/docs/lang/nodejs-iojs#locking-errors-eperm-eexist-tgz-lock

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,6 +15,8 @@ environment:
     - nodejs_version: "5"
 
 install:
+  - npm -g install npm@">=2.0.0"
+  - set PATH=%APPDATA%\npm;%PATH%
   - ps: Install-Product node $env:nodejs_version
   - npm install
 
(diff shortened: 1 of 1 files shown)


phated:
You should probably use ^2.0.0 so we always get version 2 and not 3.

sttk:
@phated I tried `npm@^2.0.0` and `npm@"^2.0.0"` but both failed.
To get version 2 and not 3, `npm@2` would be correct.

And the following would be the most appropriate:
```
  - IF %nodejs_version% EQU 0.10 npm -g install npm@2
  - IF %nodejs_version% EQU 0.10 set PATH=%APPDATA%\npm;%PATH%
```
(The error was raised only in version 0.10 of node.js.)

phated:
Thanks!

Issue #57, opened 2016-01-31
Apply --depth flag to --tasks-json flag output


phated:
In 1.2.0, @sttk added a `--depth` flag to gulp-cli; However, we didn't apply it to the `--tasks-json` flag to reduce the exponential growth as described in https://github.com/gulpjs/gulp/issues/1033

@sttk If you or anyone else would like to expand on the functionality you added for `--depth` and apply it to `--tasks-json`, that would be super helpful.

sttk:
@phated I'm willing to tackle this also, but I have a question.

There are cases that `--depth` cuts child tasks of series/parallel tasks. Are the json data of such series/parallel tasks useful? 

I think `--tasks-json` should be able to output in two type forms, a full tree and a tree until first child tasks which are not series/parallel tasks. And the latter form would be possible if gulp-cli can get `branch` flags in undertaker.

sttk:
However, it may be better that `--tasks` and `--tasks-json` can output same trees by `--depth`. How do you think about my idea below?

```js
(default or --depth=full)
[07:09:11] ├─┬ task0
[07:09:11] │ └─┬ <series>
[07:09:11] │   ├─┬ <parallel>
[07:09:11] │   │ ├── func1
[07:09:11] │   │ └── func2
[07:09:11] │   └── func3
[07:09:11] └─┬ task1
[07:09:11]   └─┬ <series>
[07:09:11]     └─┬ task0
[07:09:11]       └─┬ <series>
[07:09:11]         ├─┬ <parallel>
[07:09:11]         │ ├── func1
[07:09:11]         │ └── func2
[07:09:11]         └── func3
```
```js
(--depth=3)
[07:10:10] ├─┬ task0
[07:10:10] │ └─┬ <series>
[07:10:10] │   ├── <parallel>
[07:10:10] │   └── func3
[07:10:10] └─┬ task1
[07:10:10]   └─┬ <series>
[07:10:10]     └── task0

(--depth=4)
[07:11:37] ├─┬ task0
[07:11:37] │ └─┬ <series>
[07:11:37] │   ├─┬ <parallel>
[07:11:37] │   │ ├── func1
[07:11:37] │   │ └── func2
[07:11:37] │   └── func3
[07:11:37] └─┬ task1
[07:11:37]   └─┬ <series>
[07:11:37]     └─┬ task0
[07:11:37]       └── <series>
```
```js
(--depth=nonrecursive, or another name. I didn't find an appropriate name.) 
[...]

PR #75, opened 2016-04-05
Apply --depth and --compact flags to --tasks-json and --tasks

sttk:
This is a PR about the issue #57

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -59,14 +59,14 @@ gulp has very few flags to know about. All other flags are for tasks to use if n
 **--tasks-json** [path]
     Print the task dependency tree, in JSON format, for the loaded gulpfile. The [path] argument is optional, and if given writes the JSON to the path.
 
-**--depth** [number]
+**--tasks-depth** [number]
     Specify the depth of the task dependency tree to print. This flag can be used with --tasks or --tasks-json.
 
-**--compact**
+**--compact-tasks**
     Reduce the output of task dependency tree by printing only top tasks and their child tasks. This flag can be used with --tasks or --tasks-json.
 
-**--no-sort**
-    Will not sort top tasks of task dependency tree. This flag can be used with --tasks.
+**--sort-tasks**
+    Will sort top tasks of task dependency tree. This flag can be used with --tasks.
 
 **--color**
     Will force gulp and gulp plugins to display colors, even when no color support is detected.
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -60,7 +60,7 @@ gulp has very few flags to know about. All other flags are for tasks to use if n
     Print the task dependency tree, in JSON format, for the loaded gulpfile. The [path] argument is optional, and if given writes the JSON to the path.
 
 **--tasks-depth** [number]
-    Specify the depth of the task dependency tree to print. This flag can be used with --tasks or --tasks-json.
+    Specify the depth of the task dependency tree to print. This flag can be used with --tasks or --tasks-json. (This flag was named --depth before but is deprecated.)
 
 **--compact-tasks**
     Reduce the output of task dependency tree by printing only top tasks and their child tasks. This flag can be used with --tasks or --tasks-json.
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -59,6 +59,15 @@ gulp has very few flags to know about. All other flags are for tasks to use if n
 **--tasks-json** [path]
     Print the task dependency tree, in JSON format, for the loaded gulpfile. The [path] argument is optional, and if given writes the JSON to the path.
 
+**--depth** [number]
+    Specify the depth of the task dependency tree to print. This flag can be used with --tasks or --tasks-json.
+
+**--compact**
+    Reduce the output of task dependency tree by printing only top tasks and their child tasks. This flag can be used with --tasks or --tasks-json.
+
+**--no-sort**
+    Will not sort top tasks of task dependency tree. This flag can be used with --tasks.
+
 **--color**
     Will force gulp and gulp plugins to display colors, even when no color support is detected.
 
--- a/lib/shared/cliOptions.js
+++ b/lib/shared/cliOptions.js
@@ -58,6 +58,7 @@ module.exports = {
       'in JSON format, for the loaded gulpfile.'),
   },
   'tasks-depth': {
+    alias: 'depth',
     type: 'number',
     requiresArg: true,
     desc: chalk.gray(
--- a/lib/shared/cliOptions.js
+++ b/lib/shared/cliOptions.js
@@ -57,22 +57,22 @@ module.exports = {
       'Print the task dependency tree, ' +
       'in JSON format, for the loaded gulpfile.'),
   },
-  depth: {
+  'tasks-depth': {
     type: 'number',
     requiresArg: true,
     desc: chalk.gray(
       'Specify the depth of the task dependency tree.'),
   },
-  compact: {
+  'compact-tasks': {
     type: 'boolean',
     desc: chalk.gray(
       'Reduce the output of task dependency tree by printing ' +
       'only top tasks and their child tasks.'),
   },
-  'no-sort': {
+  'sort-tasks': {
     type: 'boolean',
     desc: chalk.gray(
-      'Will not sort top tasks of task dependency tree.'),
+      'Will sort top tasks of task dependency tree.'),
   },
   color: {
     type: 'boolean',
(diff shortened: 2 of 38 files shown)


sttk:
npm install EPERM errors frequently occured on appveyor.
It seems that this error can be fixed by using npm3.
So I modified appveyor.yml to install latest npm.

npm/npm#9696

sttk:
Sorry. I forgot to rename the second argument of `logTasks`.

sttk:
I've reflected the change of PR #87 for undertaker v1.0.0 and added `--no-sort` flag.

phated:
This will need to be rebased on top of the merged undertaker 1.0 fixes.

sttk:
I've rebased this on the current master branch.

phated:
@sttk This will probably need many changes to support only through the config files.

sttk:
@phated O.K. I'm going to address this from now.

sttk:
@phated I've finished reflecting the latest master branch.

phated:
@sttk thanks for rebasing.  However, I don't think we should expose the `--compact`, `--depth` or `--no-sort` options as command line flags.  Instead, could you implement this to only be configurable from `.gulp.*` files?

phated:
Sorry, I forgot that we shipped `--depth` but it looks like `--compact` and `--no-sort` are being added.  I'm not sure I want to keep expanding the flags but maybe those are a must-have for @segrey's use case?

sttk:
In his use case, gulp would be called from a program. So I think `--compact` flag is needed rather than config.

About `--depth` flag, I suppose users want to use both flag and config. For example, they usually specify depth=3 with config, and sometimes show full depth  tree with the flag.

About `--no-sort` flag, I think it may be good to be config only.

sttk:
@phated Could you define config keys of compact, depth, sort/no-sort?

phated:
@sttk if you are going to support `--compact` and `--depth` as flags, they should just be `flags.compact` and `flags.depth`; however, do we want to make them more descriptive (e.g. `--compact-tasks` and `--tasks-depth`)?  Do we really need the `sort`/`no-sort` option? I'm not sure who would want to turn sorting off.

sttk:
I also think `--compact-tasks` and `--tasks-depth` are better. These names make more clear what the flags is for.

`no-sort` was requested by @erikkemperman in the issue #50.

erikkemperman:
Yeah, and actually I might even argue the default should be to respect the order of tasks in my gulpfile, where I tend to try and find some sensible structure and keep related tasks near one another. Then I'd like to see the same order when listing the tasks. But maybe that's just me?

phated:
@erikkemperman I think that's a sane default, which I believe is equivalent to "no sort" because I just iterate over the object (it's not guaranteed to be in insert order, but highly likely).

erikkemperman:
Maybe I misunderstand, I thought the proposal here is to do alphabetical sorting by default, and have a "no-sort" option to get file-order. I would prefer the other way around. But again, maybe that's just me and anyway it's not a big deal :-)

phated:
@erikkemperman the original implementation sorted by default.  This PR is adding the option to "disable" that but I'm suggesting we invert that assumption (remove sorting by default and add a `--sort-tasks` flag).

erikkemperman:
Excellent, that's what I would expect :-)

phated:
@sttk are you still planning to update this PR based on the discussion?

sttk:
@phated Sorry that I'm late to update.

I implemented and I'm testing about this PR from now.

I'm changing the flag names to: `--compact-tasks`, `--tasks-depth`, `--sort-tasks`, and adding the configs: `flags.compactTasks`, `flags.tasksDepth`, `flags.sortTasks` in accordance with the flag names.

Please let me know if there are flags or configs that should be renamed or removed,

phated:
@sttk I think all of those flags & configs are good.

sttk:
@phated I've rebased and updated.

docs/CLI.md
```diff
@@ -59,6 +59,15 @@ gulp has very few flags to know about. All other flags are for tasks to use if n
 **--tasks-json** [path]
     Print the task dependency tree, in JSON format, for the loaded gulpfile. The [path] argument is optional, and if given writes the JSON to the path.
 
+**--depth** [number]
+    Specify the depth of the task dependency tree to print. This flag can be used with --tasks or --tasks-json.
```
tkellen: This should be tasks-depth, right?

docs/CLI.md
```diff
@@ -59,6 +59,15 @@ gulp has very few flags to know about. All other flags are for tasks to use if n
...
+
+**--compact**
```
tkellen: This should be compact-tasks, right?

sttk:
@tkellen Thanks! I've modified what you pointed out.

lib/shared/cliOptions.js
```diff
@@ -47,12 +47,6 @@ module.exports = {
     desc: chalk.gray(
       'Print the task dependency tree for the loaded gulpfile.'),
   },
-  depth: {
```
phated: This is a breaking change. Can we leave it in as an alias? Does yargs support aliases in the definition?

phated: If we alias this, it'll need to be documented and noted as deprecated.

lib/versioned/^3.7.0/index.js
```diff
@@ -34,20 +37,56 @@ function execute(opts, env, config) {
...
         tree.label = config.description;
       } else {
         tree.label = 'Tasks for ' + chalk.magenta(tildify(env.configPath));
       }
-      return logTasks(tree, opts.depth, function(task) {
+      return logTasks(tree, opts, function(task) {
         return gulpInst.tasks[task].fn;
       });
     }
+    if (opts.tasksJson) {
```
phated: Can this addition be moved to a separate commit/PR?  I'm glad we are adding it but I want it to show up in the changelog properly.

test/expected/flags-tasks-depth4.txt
```diff
@@ -0,0 +1,17 @@
...
+│   └── func2
+├─┬ taskB
+│ └─┬ <parallel>
+│   ├── func3
+│   └─┬ taskC
+│     └── <series>
+└─┬ default
+  └─┬ <parallel>
+    ├─┬ taskC
+    │ └── <series>
```
phated: Is this considered part of the depth?

lib/shared/cliOptions.js
sttk: I rethought about flag names that `--depth`, `--compact` and `--sort` are better than `--tasks-depth`, `--compact-tasks` and `--sort-tasks`. When we use these flags, we type a command like `gulp --tasks --tasks-depth`. The word "tasks" is redundant for users, I think.


I think good about the config names, but it is possible to group these flags like `flags.tasks.depth`, `flags.tasks.sort`, `flags.tasks.compact` (in .gulp.json, `{ flags: { tasks: { depth: 3, sort: true, compact: true } } }`).

What do you think?

lib/versioned/^3.7.0/index.js
sttk: I got it. I'll remove this part from this PR.

test/expected/flags-tasks-depth4.txt
sttk: Yes. `depth` flag cuts deeper tasks forcely.

lib/shared/cliOptions.js
phated: I'm fine with being verbose on the flag names but I think we just need to alias `depth` because we've previously defined it and removing it would be a breaking change (and I'd have to publish as 2.0.0)

test/expected/flags-tasks-depth4.txt
phated: Oh, okay

lib/shared/cliOptions.js
sttk: I got it. I'll add the alias `depth`.

sttk:
I've rebased and modified.

phated:
Looks good, great work again @sttk 

I'm going to squash this, feel free to submit the `--tasks-json` feature for V3 afterwards.

sttk:
@phated Thank you for merging.

phated:
@sttk No problem.  I'm noticing some timeout problems (for a couple tests) on Appveyor.  Did you run into that at all?

sttk:
@phated I sometimes encounter timeout problems on Appveyor, too. And in that case, I try test 2 or 3 times. I don't know this workaround.

PR #76, opened 2016-05-14
Update CI

adam-moss:
Update Travis with the latest version of node, major version releases of node, and iojs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,11 @@
 sudo: false
 language: node_js
 node_js:
-  - "stable"
-  - "4"
+  - "node"
+  - "iojs"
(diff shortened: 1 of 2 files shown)


adam-moss:
The `Reflect` leak is caused by the outdated version of `lab`.  PR #77 would resolve this, however that in term breaks support for 0.10 and 0.12.

phated:
Thanks for sending this.  We are currently in the process of switching to mocha + expect for testing because lab no longer aligns with our perspectives.

Our only supported platforms will be node 0.10, 0.12, 4, 5 and 6; we will not be supporting specific version of node like 4.2 nor iojs releases.

PR #79, opened 2016-06-09
Args implementation

leo:
After taking a look on the current implementation of the auto-generated usage information and some other things related to command line arguments within gulp-cli, I've decided that it might be a good idea to set up a real, awesome usage information by adding support for [args](https://github.com/leo/args).

A nice side-effect of this whole PR is that the code is a lot cleaner than before because args is doing a lot of abstraction on top of minimist (not like yargs does it). Here's an example:

screen shot 2016-06-09 at 17 05 12

As you can see, there's also a tiny bug at the end of the outputted usage information. I have no idea where it comes from and I've already spent hours on searching args for that error, but it looks like there's something within gulp-cli that simply cuts of the output. Maybe someone in here has an idea?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,9 @@
 sudo: false
 language: node_js
 node_js:
-  - "stable"
-  - "4"
-  - "0.12"
-  - "0.10"
+  - 6
+  - 5
+  - 4
 before_script:
   - sed -i 's/4.0.0-alpha.1/4.0.0/g' node_modules/gulp/package.json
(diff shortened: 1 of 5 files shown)


terinjokes:
There's still quite a bit of users using node.js versions 0.10 and 0.12, as a result I would be hesitant with dropping support for them.

phated:
Sticking with yargs. Sorry.

leo:
Ok... ¯\_(ツ)_/¯

PR #81, opened 2016-06-11
Remove shebang and x mode from index.js (fix #80); fix tests

### LINKED ISSUE #80: Should index.js be executable?
Building a tool with separate core and cli, I was looking at gulp and gulp-cli (4.0 branches) for an example. I noticed that `gulp-cli/index.js` has a shebang line and executable mode, but I can't figure out why...

The script exports a function, running it with `./index.js` doesn't really do anything, it seems to me. Is there something clever going on here, or is this maybe a leftover of an earlier setup? It looks like it would have made sense before these:

https://github.com/gulpjs/gulp/commit/95ce47adc59bc882370e21c48557a766efc62998

https://github.com/gulpjs/gulp-cli/commit/6b2c2ef8827980af5f9c3b0b107173aa70c484ca

I was also wondering why gulp has its own executable `bin/gulp.js`, which just requires gulp-cli and runs the exported function -- but now I suspect that this is in case users do `npm install -g gulp` (instead of installing only gulp-cli globally as recommended in the getting started guide). Have I got that right?

erikkemperman:
Being pedantic I guess, but might as well follow up on https://github.com/gulpjs/gulp-cli/issues/80.

Noticed that Travis failed, tracked that down to [this commit](https://github.com/gulpjs/undertaker/commit/cb2bcfea11f8f0941cf9ed3b07f04a53b115a3e8) in Undertaker, and adapted the test here accordingly.


erikkemperman:
Hm, I just noticed @sttk 's PR [#75](https://github.com/gulpjs/gulp-cli/pull/75) which addresses the same Undertaker change, but is aimed at gulp-cli/master. Turns out the change I made to ` test/expected/flags-tasks-json.json` for this PR to the 4.0 branch is already in master.

So now I'm confused, should I have submitted this to master?

phated:
@erikkemperman The 4.0 branch is old and outdated.  Master has everything from 4.0 branch plus lots more.  I have left it around because people wrote tutorials that mentioned it.  It seems that I should delete it now.

phated:
Deleted.  Please submit to master.  Thanks

erikkemperman:
Thanks for clearing that up, I've re-submitted to master.

PR #82, opened 2016-06-14
Remove shebang and x mode from index.js (fix #80)

### LINKED ISSUE #80: Should index.js be executable?

erikkemperman:
Re-submitted to master, as per https://github.com/gulpjs/gulp-cli/pull/81#issuecomment-225708541

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,5 @@
-#!/usr/bin/env node
-
 'use strict';
+
 var fs = require('fs');
 var path = require('path');
 var log = require('gulplog');


phated:
Thanks!

phated:
Published as 1.2.2

PR #87, opened 2016-06-30
Adapt to undertaker v1.0.0

sttk:
This PR is for adapting to both v1.0.0 and previous versions of `undertaker`.
(And this adds a test case for 100% coverage.)


appveyor.yml
```diff
@@ -15,8 +15,8 @@ environment:
     - nodejs_version: "5"
 
 install:
-  - IF %nodejs_version% EQU 0.10 npm -g install npm@2
-  - IF %nodejs_version% EQU 0.10 set PATH=%APPDATA%\npm;%PATH%
+  - npm -g install npm@latest
```
phated: are you sure you want to be using npm 3?  It's so much slower, and appveyor is already extremely slow.

lib/shared/log/tasks.js
```diff
@@ -13,7 +13,23 @@ function isObject(val) {
   return typeof val === 'object' && !Array.isArray(val);
 }
 
+function wrapGetTask(getTask) {
```
phated: this probably shouldn't be in the shared module and should instead go in the `getTask` methods of alpha.2 and 4.0

lib/shared/log/tasks.js
```diff
@@ -13,7 +13,23 @@ function isObject(val) {
...
+  return function(name) {
+    return createTask(getTask(name));
+  };
+}
+
+function createTask(task) {
+  var unwrapped = task.unwrap ? task.unwrap() : {};
```
phated: 1. try to read `task.description` before unwrapping
2. if it is empty, check for `typeof task.unwrap === 'function'`, unwrap and check for `unwrapped.description`

package.json
```diff
@@ -27,7 +27,7 @@
     "lint": "eslint . && jscs index.js bin/ lib/ test/",
     "prepublish": "marked-man --name gulp docs/CLI.md > gulp.1",
     "pretest": "npm run lint",
-    "test": "lab test/*.js -cv",
+    "test": "lab test/*.js -cv -I Reflect",
```
phated: yeah, this is annoying.  These tests will need to be switch to mocha like all the other repositories.  This is a fine fix for now.

appveyor.yml
sttk: npm install EPERM errors frequently occured on appveyor, and it seems that this error can be fixed by using npm3.

ref. https://github.com/npm/npm/issues/9696

lib/shared/log/tasks.js
sttk: I got it. I'll modify so.

lib/shared/log/tasks.js
sttk: I got it. I'll modify so.

lib/shared/log/tasks.js
sttk: Which do you think best to check if `task.description` is not empty (or specified)?

```js
 1. if (task.description) ...
 2. if (task.description != null) ...
 3. if (typeof task.description === 'string') ...
 4. if ('description' in task) ...
```

phated: I think `typeof task.description === 'string'` is best because we don't handle any other cases.

sttk:
Sorry. This push is a mistake. Please wait a little more.

lib/shared/log/tasks.js
sttk: Thanks. I got it. I'll do the same by using `lodash`

sttk:
I've finished modifying the codes.
I used `lodash` to check the properties and modified the same check in `tasks.js`, too.
In addition, I've added node 6 to the test target.

lib/shared/log/tasks.js
```diff
@@ -3,15 +3,11 @@
...
-var sortBy = require('lodash.sortby');
 
-function isString(val) {
-  return typeof val === 'string';
-}
-
-function isObject(val) {
-  return typeof val === 'object' && !Array.isArray(val);
-}
+var _ = require('lodash');
```
phated: I think people were trying to transition all the gulp dependencies to use `lodash.*` packages, so this should require `lodash.sortby`, `lodash.isstring` and `lodash.isplainobject`

lib/versioned/^4.0.0/log/getTask.js
```diff
@@ -0,0 +1,39 @@
...
+
+var _ = require('lodash');
+var isString = _.isString;
+var isObject = _.isPlainObject;
+var isFunction = _.isFunction;
+
+function getTask(gulpInst) {
+  return function(name) {
+    var task = gulpInst.task(name);
+    var func = isFunction(task.unwrap) ? task.unwrap() : {};
```
phated: this can be moved to the `getDescription` and `getFlags` functions.  It will end up calling `unwrap` twice in some scenarios but that shouldn't be much overhead and it avoids it if the properties are on the wrapper

package.json
```diff
@@ -37,7 +37,7 @@
     "gulplog": "^1.0.0",
     "interpret": "^1.0.0",
     "liftoff": "^2.1.0",
-    "lodash.sortby": "^4.0.1",
+    "lodash": "^4.13.1",
```
phated: avoid using the full lodash package, instead use the small modules

test/fixtures/flags-tasks-with-nowrap.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
+
+var gulp = require('gulp');
+
+gulp.task('task1', function() {});
+gulp.task('task1').description = 'Description for gulp.task("task1")';
+
+gulp.task('task2', function() {});
+if (!gulp.task('task2').unwrap) {
```
phated: instead of all this, can't you just use a named function?

test/fixtures/flags-tasks-with-nowrap.js
```diff
@@ -0,0 +1,40 @@
...
+  gulp.task('task2').unwrap = function() {
+    return fn2;
+  };
+}
+gulp.task('task2').unwrap()
+  .description = 'Description for gulp.task("task2").unwrap()';
+
+gulp.task('task3', function() {});
+gulp.task('task3', function() {});
+if (!gulp.task('task3').unwrap) {
```
phated: instead of all this, can't you just use a named function?

test/fixtures/flags-tasks-with-nowrap.js
```diff
@@ -0,0 +1,40 @@
...
+if (!gulp.task('task2').unwrap) {
+  var fn2 = function() {};
+  gulp.task('task2').unwrap = function() {
+    return fn2;
+  };
+}
+gulp.task('task2').unwrap()
+  .description = 'Description for gulp.task("task2").unwrap()';
+
+gulp.task('task3', function() {});
```
phated: why is `task3` set twice?

test/fixtures/flags-tasks-with-nowrap.js
```diff
@@ -0,0 +1,40 @@
...
+    return fn3;
+  };
+}
+gulp.task('task3')
+  .description = 'Use gulp.task("task3").description preferentially';
+gulp.task('task3').unwrap().description = 'This description should not output';
+
+gulp.task('no-desc', function() {});
+gulp.task('no-desc', function() {});
+if (!gulp.task('no-desc').unwrap) {
```
phated: instead of all this, can't you just use a named function?

test/taskTree.js
```diff
@@ -48,4 +48,58 @@ lab.experiment('taskTree()', function() {
     code.expect(taskTree(tasks)).to.deep.equal(expectTree);
     done();
   });
+
+  lab.test('processes children recursively.', function(done) {
```
phated: where did this test come from?

phated:
@sttk sorry it took me so long to review.  I've commented on some things I noticed.

sttk:
@phated Don't worry. I know you are busy with other projects. Thank you for reviewing.

test/fixtures/flags-tasks-with-nowrap.js
sttk: The purpose of this test is that `gulp-cli` supports both the old way, `description` is gotten by `gulp.task(name)`, and the new way, it is gotten by `gulp.task(name).unwrap()`.

This cannot be controlled by setting using named function.

And getting task tree correctly by named functions is tested by `flags-tasks-with-desc.js`.

test/fixtures/flags-tasks-with-nowrap.js
sttk: It might be my mistake or I couldn't remember the reaseon. I'll remove one line of them and one more line about 'no-desc'.

test/taskTree.js
sttk: This test is what I forgot to include in previous PR and for coverage 100%.

sttk:
@phated I've modified what you pointed out. And I've merged the tests for the flag `--tasks` in one.

phated:
Published as 1.2.2

phated:
Excellent work @sttk!  Really glad you caught this oversight in my undertaker changes.

Issue #69, opened 2016-04-05
Implement .gulprc support


phated:
Implement through use of https://github.com/js-cli/js-liftoff/issues/8 which has outstanding PR at https://github.com/js-cli/js-liftoff/pull/67

phated:
Support for this in Liftoff was just published as 2.3.0

@sttk would you like to implement it for gulp?

sttk:
@phated Yes, I'd like to implement it. 

I want to use key chain style for gulprc as below, but it is not popular as a configuration file. Can I use this style and the module [customerizer](https://github.com/sttk/customizer.git) for this?                                       

```js
cli.flags.cwd = 'path/of/cwd';
cli.flags.logLevel = 4; 

colors.tasks.tree.task.name = s => chalk.bold(chalk.blue(s));
colors.tasks.tree.task.description = chalk.black;
```

phated:
@sttk I don't think the configuration file needs to be that complex.  I also think that attempting the implementation with the most simple thing first is the correct approach (so don't start with colors).  I think the easiest thing to support would be a top level `description` property as outlined in #70.  The `.gulprc` would then look like:
```json
{
  "description": "My sample description for gulpfile"
}
```
It could also be a node module that used `module.exports = ...` because we will just be using `require` to load the file.

sttk:
@phated I got it. And I agree that we should start with the most simple configuration.

> ... we will just be using `require` to load the file.

`require` and `rechoir` can't get the extension `"rc"` from filename `".gulprc"`. So they loads the file as a js file.

For autoloading `.gulprc` with `rechoir`, we need to do follwing two things:

1. Create a new loader for JSON file. This code would do a thing like `module.exports = JSON.parse(fs.readFileSync(fp, opts))` and something similar to [module.load](https://github.com/nodejs/node/blob/master/lib/module.js#L449) except resolving extension.

2. Change `extension` module in `rechoir` to enable to get the extension `"rc"` from the filename `".gulprc"`. This would be made possible by seeking a matching tail in `rechoir`'s argument `config`.

sttk:
@phated Can I change `rechoir` as above?

phated:
Oh, you are right that `.gulprc` won't work out of the box with require. Let's just support `.gulp.js`, etc (all the extensions we already support).

sttk:
I got it. I'll implement so.

sttk:
[...]

PR #90, opened 2016-08-10
Implement .gulprc support

sttk:
This is a PR about the issue #69. This make possible to operate configuration files as below:

* Load and merge configuration files in home directory, project directory, current work directory if exists.
* Print configurations in JSON format by `--output-config` flag.
* Load and merge configuration files by `--config` or `-c` flag.
*  Ignore configuration files by `--ignore-config` flag, except files specified by `--config` flag.

In addition, this PR adds some useful test tools in `test/tools` directory (which are tested with files in `test/tools/test` directory):

* `run-gulp` : creates and executes a command line string which changes directories and runs gulp with options (and can print the command string, error, stdout and stderr)
* `erase-time` : erases timestamp from each lines in output text.
* `skip-lines` : removes first specified number of lines in output text.


docs/CLI.md
```diff
@@ -61,3 +61,12 @@ gulp has very few flags to know about. All other flags are for tasks to use if n
...
 **--log-level**, **-L**
     Set the loglevel. -L for least verbose and -LLLL for most verbose. -LLL is default.
+
+**--config**, **-c**
+    Manually set path of configuration file.
+
+**--ignore-config**
+    Ignore configuration files except files specified by --config flag.
+
+**--output-config**
```
phated: Who/what would use this flag?  What is the purpose and benefit?

docs/CLI.md
```diff
@@ -61,3 +61,12 @@ gulp has very few flags to know about. All other flags are for tasks to use if n
 
 **--log-level**, **-L**
     Set the loglevel. -L for least verbose and -LLLL for most verbose. -LLL is default.
+
+**--config**, **-c**
+    Manually set path of configuration file.
+
+**--ignore-config**
```
phated: Do we need this right away?  The config files are an advanced feature that is opt-in by placing the files on the filesystem.  Maybe this shouldn't be added until someone needs it.

docs/CLI.md
```diff
@@ -61,3 +61,12 @@ gulp has very few flags to know about. All other flags are for tasks to use if n
 
 **--log-level**, **-L**
     Set the loglevel. -L for least verbose and -LLLL for most verbose. -LLL is default.
+
+**--config**, **-c**
```
phated: Not sure about having this option either.  I was expecting the config files to be static (always named `.gulp.{ext}`).  Why do we want to make this configurable?

index.js
```diff
@@ -34,6 +39,22 @@ var cli = new Liftoff({
   completions: completion,
   extensions: interpret.jsVariants,
   v8flags: v8flags,
+  configFiles: {
+    '.gulprc': {
```
phated: just name this `.gulp`

index.js
```diff
@@ -34,6 +39,22 @@ var cli = new Liftoff({
...
+      home: {
+        path: '~',
+        extensions: interpret.extensions,
```
phated: we only want to support the `jsVariants` so we should allow this to be defaulted by the top-level `extensions`.  No need to specify again.

index.js
```diff
@@ -34,6 +39,22 @@ var cli = new Liftoff({
...
+      },
+      project: {
+        path: '.',
+        extensions: interpret.extensions,
```
phated: we only want to support the `jsVariants` so we should allow this to be defaulted by the top-level `extensions`.  No need to specify again.

index.js
```diff
@@ -34,6 +39,22 @@ var cli = new Liftoff({
...
+      },
+      cwd: {
+        path: process.env.INIT_CWD,
+        extensions: interpret.extensions,
```
phated: we only want to support the `jsVariants` so we should allow this to be defaulted by the top-level `extensions`.  No need to specify again.

phated: Do we want to support `cwd`?  I think that might start to confuse people

index.js
```diff
@@ -121,6 +152,26 @@ function handleArguments(env) {
     exit(1);
   }
 
+  customizeConfig(['home', 'project', 'cwd'].map(getConfigFile));
```
phated: this is mapping but not assigning to a variable

lib/shared/config/customize.js
```diff
@@ -0,0 +1,22 @@
...
+  if (lastLoadedFile === file) {
+    return false;
+  }
+  lastLoadedFile = file;
+  return true;
+}
+
+module.exports = function(configFiles) {
+  configFiles.filter(isString).filter(isNotLastLoaded).forEach(function(file) {
+    merge(config, require(file));
```
phated: you wouldn't want to merge onto the `config` singleton

lib/versioned/^3.7.0/index.js
```diff
@@ -38,7 +41,8 @@ function execute(opts, env) {
     }
     if (opts.tasks) {
       var tree = taskTree(gulpInst.tasks);
-      tree.label = 'Tasks for ' + chalk.magenta(tildify(env.configPath));
+      tree.label = isString(config.description) ? config.description :
```
phated: It's not enough to check `isString`.  We don't want to print an empty string description

lib/versioned/^4.0.0-alpha.1/index.js
```diff
@@ -42,10 +45,10 @@ function execute(opts, env) {
       return logTasksSimple(gulpInst.tree());
     }
     if (opts.tasks) {
-      var tree = {
-        label: 'Tasks for ' + chalk.magenta(tildify(env.configPath)),
-        nodes: gulpInst.tree({ deep: true }),
-      };
+      var tree = {};
+      tree.label = isString(config.description) ? config.description :
```
phated: It's not enough to check `isString`.  We don't want to print an empty string description

lib/versioned/^4.0.0-alpha.2/index.js
```diff
@@ -48,13 +51,15 @@ function execute(opts, env) {
     }
     if (opts.tasks) {
       tree = gulpInst.tree({ deep: true });
-      tree.label = 'Tasks for ' + chalk.magenta(tildify(env.configPath));
+      tree.label = isString(config.description) ? config.description :
```
phated: It's not enough to check `isString`.  We don't want to print an empty string description

lib/versioned/^4.0.0-alpha.2/index.js
```diff
@@ -48,13 +51,15 @@ function execute(opts, env) {
...
+        'Tasks for ' + chalk.magenta(tildify(env.configPath));
 
       return logTasks(tree, opts.depth, getTask(gulpInst));
     }
     if (opts.tasksJson) {
       tree = gulpInst.tree({ deep: true });
-      tree.label = 'Tasks for ' + tildify(env.configPath);
+      tree.label = isString(config.description) ? config.description :
```
phated: It's not enough to check `isString`.  We don't want to print an empty string description

lib/versioned/^4.0.0/index.js
```diff
@@ -48,13 +51,15 @@ function execute(opts, env) {
     }
     if (opts.tasks) {
       tree = gulpInst.tree({ deep: true });
-      tree.label = 'Tasks for ' + chalk.magenta(tildify(env.configPath));
+      tree.label = isString(config.description) ? config.description :
```
phated: It's not enough to check `isString`.  We don't want to print an empty string description

lib/versioned/^4.0.0/index.js
```diff
@@ -48,13 +51,15 @@ function execute(opts, env) {
...
+        'Tasks for ' + chalk.magenta(tildify(env.configPath));
 
       return logTasks(tree, opts.depth, getTask(gulpInst));
     }
     if (opts.tasksJson) {
       tree = gulpInst.tree({ deep: true });
-      tree.label = 'Tasks for ' + tildify(env.configPath);
+      tree.label = isString(config.description) ? config.description :
```
phated: It's not enough to check `isString`.  We don't want to print an empty string description

docs/CLI.md
sttk: By this flag, users can check the current configurations and get the list of the customizable configurations. Since the configuration set is the result of merging multiple files, it would be useful to check the result. And it would be useful for sharing someone's configurations, for generating single configuration files from multiple configuration files by combining it with `--config` flag, and so on.

docs/CLI.md
sttk: Certainly, we don't need it right away. But in the future, when a lot of configurations can be customized by the config files, there would be some cases that an user want to off his own configurations, e.g. for testing configuration files temporarily, for trying or debugging other person's code not to be affected by them, etc.

docs/CLI.md
sttk: This flag is useful for using gulp while switching some configuration files, for applying some configurations temporarily, for testing configuration files, etc.

These three flags above are not necessary until someone needs them, as you said. We can remove these features because it is easy to add them again.

index.js
sttk: I got it.

index.js
sttk: But `jsVariants` doesn't contain `'.json'`( [js-interpret/index.js#L106](https://github.com/js-cli/js-interpret/blob/master/index.js#L106) )

index.js
sttk: Yes, I think so. For example, there were some issues which wanted to specify the path of `gulpfile.js` in a gulprc file. I suppose the cases that a directory of a config file is different from a directory of a gulpfile.js like that.

sttk: But it would be good to add 'cwd' latar until it is needed.

index.js
sttk: The purpose of this code is to pass config. properties to `customizeConfig` function in the order of the keys `home`, `project` and `cwd`, because object properties order is not guaranteed.

lib/shared/config/customize.js
sttk: I suppose the configuration will have nested properties in the future.

lib/versioned/^3.7.0/index.js
sttk: I got it. I'll add `!config.description` to this condition.

docs/CLI.md
phated: That's an interesting use case.  I think each of these should come later and as separate PRs.

docs/CLI.md
phated: I think each of these should come later and as separate PRs.

docs/CLI.md
phated: I think each of these should come later and as separate PRs.

index.js
phated: Good catch!  I actually want to write my configs as `.gulp.yml` which also isn't supported in `jsVariants`.  Please leave it as `interpret.extensions`.

index.js
phated: If we support `cwd`, I'm not sure it should be using `INIT_CWD` because someone could specify `--cwd` option that changes the `cwd`.  Thoughts?

index.js
phated: This code seems a little convoluted for what it is doing.  Maybe you can refactor to make it more clear.

lib/shared/config/customize.js
phated: You'd want to merge into an empty object so you don't mutate the module (`config`) itself.

docs/CLI.md
sttk: I got it. I'll remove these features in this PR and add them later when they are needed.

index.js
sttk: I mistook. When `gulpfile.js` is not found, `env.cwd` become same with `process.env.INIT_CWD`. So this `cwd` property is not needed. I'll remove this property.
(And It would be better to rename the key name `project` to `cwd`.)

index.js
sttk: Is it better to use `sortBy` as follows?

```js
customizeConfig(sortBy(env.configFiles['.gulp'], ['home', 'project', 'cwd']));
```

lib/shared/config/customize.js
sttk: I think `config` has to be singleton, because it has to be accessible for every modules in `gulp-cli`.

Or should `config` be passed to `lib/versioned/*/index.js` as an argument?

phated: Yeah, it should be passed but I'm not sure if it should be another argument or merged with `opts` or `env` that are already passed.  Eventually we will merge the flags with `opts`.  What do you think?

index.js
phated: I think it's better to have the `sortBy` or `map` call be assigned to a variable and to make `customizeConfig` non-mutating with its return value assigned to another variable.  Naming variables for each step usually clarifies the code a lot.

index.js
phated: Sounds good to me.

lib/shared/config/customize.js
sttk: I think merging with `opts` is good! First I thought the config appears anywhere in gulp-cli and customizes its behaviors directly. But if the `opts` is merged with the config (or a new object is merged with the `opts` and the config) and gulp-cli uses only the `opts` (or the new object) for configuring its behaviors, it seems neater.

And I think this merging should be done at the top of `handleArguments` in `index.js`.

sttk:
@phated I've modified the codes.

phated:
@sttk this looks great!  Thank you so much for the work on this feature (in Liftoff and here).  I'll merge and add some docs before I publish.

phated:
`sortBy` wasn't actually doing what you thought so I'm altering it to be a `map`

phated:
Considering shipping this as 1.3.0 and pushing everything else to a 1.4 milestone.

sttk:
> sortBy wasn't actually doing what you thought so I'm altering it to be a map

I see. Thanks for altering and merging.

Issue #92, opened 2016-08-23
Restrict config properties that are merged into opts from .gulp.* files


phated:
@sttk I created this issue based on https://github.com/gulpjs/gulp-cli/issues/71#issuecomment-241882995

I'd like to keep the top-level properties restricted so they don't conflict in the future. e.g. a `--description` flag that does something different than the configuration `description` property.

Do you agree?

sttk:
@phated Sorry for my thoughtless. Certainly cli flags should be distinguished from config properties. I agree.

sttk:
I'd like to change this to use the specification object like `lib/shared/cliOptions.js` for cli flags. The implementation image is as follows:

```js
var configSpecs = require('lib/shared/configs/')
var checkConfig = require('lib/shared/configs/checkConfig')
  ...
  // The target merged with configs is an empty object.
  var config = {};

  var configFilePaths = ['home', 'cwd'].map(function(prop) {
    return env.configFiles['.gulp'][prop];
  });
  configFilePaths.filter(isString).forEach(function(filePath) {
    // The target object is merged with results of checks with the spec. object.
    merge(config, checkConfig(require(filePath), configSpecs));
  });
  ...

  // The target is marged with flags in `opts` specified in the spec object.
  merge(config, checkConfig(opts, lodash.filter(configSpecs, { cliFlag: true })));
```

And I think it's good to pass the config object to versioned `index.js` (`lib/versioned/*/index.js`) instead of `opts` in the future, but it will now be passed the config as another argument temporarily.

```js
  require(path.join(__dirname, '/lib/versioned/', range, '/'))(opts, env, config);
```

What do you think about this?

phated:
@sttk I think it is fine to continue to merge with `opts` but I don't think `.gulp.*` and `opts` should have the same shape.  I'm fine with adding helper methods.  I think we need to do something similar to `lodash.pick` but for deep properties, like `pick(require(filePath), ['description', 'flags.continue', 'flags.silent', 'flags.logLevel', ...])`. Thoughts?

sttk:
@phated 
>  I think we need to do something similar to lodash.pick but for deep properties, ...

Yes, we would need something like that. And there are `lodash.get/set` to access nested properies of an object. We would be able to use them in the helper methods.

phated:
It would be easiest to have an array of paths but I wonder if a config object that we merge into is better.

sttk:
I've implemented for this issue and sended it to PR #94.

PR #94, opened 2016-09-04
Restrict config properties (Implementation for Issue #92)

sttk:
This PR is for the issue #92.

 I only implemented about the properties: `description`, `flags.help`, and `flags.gulpfile` first. 

The points that I modified from #90 are as follows:
1. Config properties by cli-flags take precedence over them by config files.
   
   I made a mistake  in the previous implementation to merge properties of config files over `opts` properties, so config files took precedence over command line flags. I modified it. 
   
   Since it would be confusing to control to change or not `opts` properties at the time of merging, the current code evacuates `opts` properties specified by cli-flags before merging, merge with config by config files, and merge with the evacuated properties at the end.
2. Add an object `disp` for display settings.
   
   This object is used for outputing description for gulpfile and will be used for various display configurations.
   
   For this object, the api of the versioned cli modules in `lib/versioned/*/index.js` are added 3rd argument.
3. Add validations for config properties of each config file.
   
   And if a validation fails, outputs the error and quits immediately. 
4. Add convert for config properties of each config file.
   
   Because it is needed to convert a property value into an absolute path with a base directory path if the property value is a path string.
5. Add `env` to the merged targets because `env.cwd`, `env.configPath` and `env.require` are used for configuration of cwd, gulpfile and require.


sttk:
In addition, we should consider about following points:

1. Specifying `--cwd` or `--gulpfile` affect each other. Is it OK to set `flags.cwd` or `flags.gulpfile` independently about specifying by config file?

2. `--require` could be specified multiple values. Should the value of `flags.require` be able to set a string or an array? And Should the value by config files replace or append?

3. Is it OK that a base path of `flags.cwd` and `flags.gulpfile` are a directory of a `.gulp.*` file?

index.js
```diff
@@ -169,5 +166,49 @@ function handleArguments(env) {
...
+    customizeByEachConfig(config, opts, env, disp, filePath);
+  });
+
+  merge(opts, specifiedOpts);
+  merge(env, specifiedEnv);
+}
+
+function customizeByEachConfig(config, opts, env, disp, filePath) {
+  try {
+    var config = require(filePath);
```
sttk: I  found that this line is redundant with line 182. I'll remove this line or line 182.

sttk:
I've removed the redundant code.

phated:
@sttk any idea why the tests failed on node 6?

sttk:
The cause of this error is seemed to be an accidental timeout of the test for `--completion` and it is thought that this is not related to this PR.

Actually, the same test for my repository is successful.

[https://travis-ci.org/sttk/gulp-cli/builds/157911810](https://travis-ci.org/sttk/gulp-cli/builds/157911810)

phated:
@sttk Sorry for not having a chance to fully review this yet.  I don't think `description` should be placed under a `disp` key and should remain at the top level.

sttk:
@phated `disp` is not a key of `config` or `opts` but an inner object to pass to versioned runners. Since the `description` key places under `opts` in #90 implementation, this key can be also specified by unspecific `--description` cli-flag. `disp` is for preventing it.

sttk:
@phated I've changed the customization process (using `copy-props` :smile:).
The changing points are:
 
* merged each config file separately with opts, etc.
    =>  merge all config files into one, then merge with opts, etc.

* evacuated user-specified options previously, and re-merge them at the end
    =>  exclude those options from merging.

sttk:
I recreated the codes for this PR.

sttk:
I re-created the code for this PR.

This code supports and restricts only the config properties: `description`, `flags.silent` and `flags.gulpfile` first.

One problem about the property `flag.silent` occurs. Before loading config files, which is executed at the top of a callback function of `Liftoff.launch`, configuration of logging is already done in `toConsole` function.

Tentatively, I moved `toConsole` function after the loading, and for the requiring logs, which are output before the loading, I used **fancy-log** directly .

phated:
@sttk ugh, sorry that I still haven't had a chance to review this PR.  Life has been hectic for me.  I will try to review soon.

phated:
@sttk this will need to be updated after I merge the mocha/expect changes.

phated:
Feel free to rebase/update this based on the merge of #99

sttk:
@phated I'll update this PR.

sttk:
@phated I finished updating this PR.

lib/shared/config/cli-flags.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+var yargs = require('yargs');
+var camelCase = require('lodash.camelcase');
+
+function mergeToCliFlags(cliFlags, config, cliOptions) {
+  var argv = yargs(process.argv.slice(2)).argv;
```
phated: Why are you using `yargs` again here? Don't we already have everything parsed in `index.js`?

lib/shared/config/cli-flags.js
```diff
@@ -0,0 +1,24 @@
...
+  var cfgFlags = config.flags || {};
+
+  var optNames = ['silent'];
+  optNames.filter(excludeArgsByUser).map(camelCase).forEach(copyConfig);
+
+  function excludeArgsByUser(name) {
```
phated: What is this doing?

lib/shared/config/cli-flags.js
```diff
@@ -0,0 +1,24 @@
...
+
+var yargs = require('yargs');
+var camelCase = require('lodash.camelcase');
+
+function mergeToCliFlags(cliFlags, config, cliOptions) {
+  var argv = yargs(process.argv.slice(2)).argv;
+  var cfgFlags = config.flags || {};
+
+  var optNames = ['silent'];
+  optNames.filter(excludeArgsByUser).map(camelCase).forEach(copyConfig);
```
phated: Why are we mapping this with `camelCase`?

lib/shared/config/cli-flags.js
```diff
@@ -0,0 +1,24 @@
...
+
+  function excludeArgsByUser(name) {
+    return !(argv[name] != null || argv[cliOptions[name].alias] != null);
+  }
+
+  function copyConfig(name) {
```
phated: Instead of copying the config, you should be generating a new object and mapping onto that. (likely using `.reduce`)

lib/shared/config/cli-flags.js
```diff
@@ -0,0 +1,24 @@
...
+    if (cfgFlags[name] !== undefined) {
```
phated: This should probably compare with `undefined` or `null` using `!= null`

lib/shared/config/env-flags.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+var path = require('path');
+
+function mergeToEnvFlags(envFlags, config, envOpts) {
+  var cfgFlags = config.flags || {};
+
+  if (!envOpts.configPath && cfgFlags.gulpfile) {
+    envFlags.configPath = cfgFlags.gulpfile;
```
phated: This should be creating a new object instead of mutating

lib/shared/config/env-flags.js
```diff
@@ -0,0 +1,14 @@
...
+    envFlags.configBase = path.dirname(cfgFlags.gulpfile);
```
phated: This should be creating a new object instead of mutating

lib/shared/config/loadfiles.js
```diff
@@ -0,0 +1,33 @@
...
+
+var get = require('lodash.get');
+var set = require('lodash.set');
+var isString = require('lodash.isstring');
+var configSpecs = require('./cfg-specs');
+
+function loadConfigFiles(configFilesBase, keysInOrder) {
+  var config = {};
+
+  var configFilePaths = keysInOrder.map(function(key) {
```
phated: Does `fined` not do this for us automatically?  It's okay if it doesn't but my mind is fuzzy on it.

lib/shared/config/loadfiles.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+var get = require('lodash.get');
```
phated: I'm going to be removing lodash. It's okay for us to leave it in for this implementation unless you want to find replacements

lib/shared/config/loadfiles.js
```diff
@@ -0,0 +1,33 @@
...
+
+    Object.keys(configSpecs).forEach(function(keyPath) {
+      var spec = configSpecs[keyPath];
+
+      var value = get(loadedCfg, keyPath);
+      if (value === undefined) {
+        return;
+      }
+
+      set(config, keyPath, spec.validate(value, filePath));
```
phated: What is `validate` doing in this scenario?

lib/versioned/^3.7.0/index.js
```diff
@@ -39,8 +38,8 @@ function execute(opts, env) {
     }
     if (opts.tasks) {
       var tree = taskTree(gulpInst.tasks);
-      if (opts.description && isString(opts.description)) {
-        tree.label = opts.description;
+      if (config.description) {
```
phated: Will this always exist now?

index.js
```diff
@@ -20,6 +19,10 @@ var cliVersion = require('./package.json').version;
 var getBlacklist = require('./lib/shared/getBlacklist');
 var toConsole = require('./lib/shared/log/toConsole');
 
+var loadConfigFiles = require('./lib/shared/config/loadfiles');
```
sttk: I got it.

index.js
```diff
@@ -84,26 +85,35 @@ cli.on('respawn', function(flags, child) {
...
-  cli.launch({
+  var envOpts = {
     cwd: opts.cwd,
     configPath: opts.gulpfile,
     require: opts.require,
     completion: opts.completion,
-  }, handleArguments);
+  };
+
+  cli.launch(envOpts, function(env) {
```
sttk: I got it.

index.js
```diff
@@ -84,26 +85,35 @@ cli.on('respawn', function(flags, child) {
...
+    var config;
+    try {
+      config = loadConfigFiles(env.configFiles['.gulp'], ['home', 'cwd']);
```
sttk: This exception is caused by validation of config files. Does your comment mean that outputting error and exiting should be done in `loadConfigFiles`, or it is unnecessary to validate config files?

index.js
```diff
@@ -84,26 +85,35 @@ cli.on('respawn', function(flags, child) {
...
+    } catch (e) {
+      log.error(chalk.red(e.message));
+      exit(1);
+    }
+
+    mergeToCliFlags(opts, config, cliOptions);
```
sttk: I got it.

index.js
```diff
@@ -84,26 +85,35 @@ cli.on('respawn', function(flags, child) {
...
+    mergeToEnvFlags(env, config, envOpts);
```
sttk: > Why do we pass envOpts in here?

To exclude `env` property names related to command-line arguments.

> This should be non-mutating (return a new object).

I got it.

lib/shared/config/cli-flags.js
sttk: To find flags specified by command-line arguments. `opts` in `index.js` has all flags possible.

lib/shared/config/cli-flags.js
sttk: Because property names of `cliFlags` (=`opts` in `index.js`) are camel case, though property names of `argv` (command-line flag name) are hyphenated name.

lib/shared/config/cli-flags.js
sttk: This function checks if `name` is not included in command-line arguments. 

If all flags specified by config files are reflected to `opts`, flags by command-line args are overridden. I think that specifying by command-line args should be given preference over specifying by config files.

lib/shared/config/cli-flags.js
sttk: I got it.

lib/shared/config/cli-flags.js
sttk: I got it.

lib/shared/config/env-flags.js
sttk: I got it.

lib/shared/config/env-flags.js
sttk: I got it.

lib/shared/config/loadfiles.js
sttk: If it is good to use `copy-props`, I'll replace this to it.

lib/shared/config/loadfiles.js
sttk: Properties order in objects is not guaranteed, though `Object.keys` looks like returning in definition order actually. I think it is needed to specify the order explicitly.

lib/shared/config/loadfiles.js
sttk: Validating each config value comming from config files, and convert file path (e.g. gulpfile) from relative path from `.gulp.*` to absolute path.

lib/versioned/^3.7.0/index.js
sttk: About `config` object? It will be always exists.

index.js
phated: Do we need to validate the config?  Could we just ignore and log a warning or error if it's the wrong type?

I don't like the throw/try/catch pattern and we shouldn't be exiting.

index.js
phated: > To exclude env property names related to command-line arguments.

Can you explain this further?

lib/shared/config/cli-flags.js
phated: Why does it matter the ones passed in or all possible?  If you are merging, the order should determine the overrides, correct?

lib/shared/config/cli-flags.js
phated: Keys in JSON files have to be wrapped in `"` anyway, so why don't we just require hyphenated?

lib/shared/config/cli-flags.js
phated: Can you explain further?  I'm still confused.

lib/shared/config/loadfiles.js
phated: Are we using it in other places?  I'm fine with using it since you will be able to fix anything needed.

lib/shared/config/loadfiles.js
phated: Sorry, I meant that `fined` crawls the file system and only gives us the config file that matches.  Or are we using it differently?

lib/shared/config/loadfiles.js
phated: I don't like the `validate` stuff.  I think there's a better way to do this.

lib/versioned/^3.7.0/index.js
phated: Is `config.description` always guaranteed to be `undefined` or a String?

lib/shared/config/cli-flags.js
sttk: For example, when both `--gulpfile` command line option and `flags.gulpfile` in a config file are specified, command line option should be used rather than config file flag, I think.

But merging config file flags into `opts` and `env` simply, `opts.gulpfile` and `env.modulePath` specified by `--gulpfile` are overridden by `flags.gulpfile` from a config file.

So before merging, it is needed to get rid of `opts.gulpfile` and `env.modulePath` from targets to be merged. (It is also good to get rid of `flags.gulpfile` from config file flags marging.)

index.js
sttk: I got it. I'll make display errors but not exit.

lib/shared/config/loadfiles.js
sttk: If config files are found in user home directory and project directory, both file paths are given by `fined`. In that case each config in both project dir should be given priority over each config in user home dir, I think.

Though this order might be determined by descriptions at `Liftoff` constructor, I want to specify it explicitly.

lib/versioned/^3.7.0/index.js
sttk: Yes. It will be checked and output errors if invalid.

phated:
@sttk Don't mean to pester you but do you know when you'll have a chance to work on the changes?

sttk:
@phated I'll work on the changes this weekend. Why?

phated:
No reason, I've just been really active on the project lately.  I know how hard it is to get me to review, haha!

sttk:
Since I've modified matters pointed out by you, please review.
And list flags that we will be able to specify with config files, please.

phated:
@sttk I'm terrible. Sorry!  Looking at this now.

index.js
```diff
@@ -3,14 +3,13 @@
 var fs = require('fs');
 var path = require('path');
 var log = require('gulplog');
+var fancyLog = require('fancy-log');
```
phated: I really don't like using fancy-log directly, it means that someone using `--silent` will still see the messages.  Looking into re-binding the logger after we parse the config.

index.js
```diff
@@ -65,15 +68,13 @@ if (opts.continue) {
   process.env.UNDERTAKER_SETTLE = 'true';
 }
 
-// Set up event listeners for logging.
-toConsole(log, opts);
-
 cli.on('require', function(name) {
-  log.info('Requiring external module', chalk.magenta(name));
+  fancyLog('Requiring external module', chalk.magenta(name));
```
phated: See above about fancy-log

index.js
```diff
@@ -65,15 +68,13 @@ if (opts.continue) {
...
 });
 
 cli.on('requireFail', function(name) {
-  log.error(chalk.red('Failed to load external module'), chalk.magenta(name));
+  fancyLog.error(
```
phated: See above about fancy-log

index.js
```diff
@@ -89,21 +90,22 @@ function run() {
...
     completion: opts.completion,
-  }, handleArguments);
+  }, configureAndInvoke);
 }
 
 module.exports = run;
 
-// The actual logic
-function handleArguments(env) {
+function configureAndInvoke(env) {
```
phated: Is this just split out due to the linting complaining about number of statements?  I'd be happier increasing the limit and putting it inside `handleArguments`.

lib/shared/config/convert-config.js
```diff
@@ -0,0 +1,45 @@
+'use strict';
+
+var path = require('path');
+var logger = require('fancy-log');
```
phated: See above about fancy-log

lib/shared/config/convert-config.js
```diff
@@ -0,0 +1,45 @@
...
+var chalk = require('chalk');
+
+module.exports = {
+
+  description: function(value, configFile) {
```
phated: Do we really need the validations on these?  I know it would make it easier to trace but I think it's overkill for a feature this advanced.  Removing this would reduce the code by **a lot** and we could add it back in the future if we find we need it.

lib/shared/config/env-flags.js
```diff
@@ -0,0 +1,23 @@
...
+var copyProps = require('copy-props');
+
+var fromto = {
+  configPath: 'flags.gulpfile',
+  configBase: 'flags.gulpfile',
+};
+
+function mergeConfigToEnvFlags(env, config) {
+  env = copyProps(env, {});
+  return copyProps(env, config, fromto, convert, true);
```
phated: Why are you reversing this? Can't you just write it in the reverse order?

lib/shared/config/convert-config.js
phated: Also, nothing should be required.

lib/shared/config/env-flags.js
phated: Only the `fromto` values should be copied.

lib/shared/config/cli-flags.js
```diff
@@ -0,0 +1,14 @@
...
+
+var copyProps = require('copy-props');
+
+var fromto = {
+  'flags.silent': 'silent',
+};
+
+function mergeConfigToCliFlags(opt, config) {
+  opt = copyProps(opt, {});
+  return copyProps(config, opt, fromto);
```
phated: Only the `fromto` values should be copied.

lib/shared/config/exclude-cli-args.js
```diff
@@ -0,0 +1,34 @@
+'use struct';
+
+var yargs = require('yargs');
+var camelCase = require('camelcase');
+var copyProps = require('copy-props');
+var cliOptions = require('../cliOptions');
+
+function excludeCliArgs(config) {
```
phated: I still don't like this.  If we are doing the property copying correctly (only copy files that exist and only those that we expect), we shouldn't need this.

lib/shared/config/load-files.js
```diff
@@ -0,0 +1,33 @@
...
+      }
+      var convert = convertConfig[name];
+      if (typeof convert !== 'function') {
+        return value;
+      }
+      return convert(value, filePath);
+    });
+  });
+
+  config = excludeCliArgs(config);
```
phated: This shouldn't be needed. See above

lib/shared/config/load-files.js
```diff
@@ -0,0 +1,33 @@
...
+    var filePath = configFilesBase[fileTitle];
+    if (!filePath) {
+      return;
+    }
+
+    copyProps(require(filePath), config, function(value, name) {
+      if (value == null) {
+        return undefined;
+      }
+      var convert = convertConfig[name];
```
phated: This validation shouldn't be needed.

lib/shared/config/load-files.js
```diff
@@ -0,0 +1,33 @@
...
+      if (typeof convert !== 'function') {
+        return value;
+      }
+      return convert(value, filePath);
```
phated: This validation shouldn't be needed.

lib/shared/config/load-files.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+var copyProps = require('copy-props');
+var convertConfig = require('./convert-config');
+var excludeCliArgs = require('./exclude-cli-args');
+
+function loadConfigFiles(configFilesBase, configFileTitles) {
```
phated: `configFileTitles` should be named `configFileOrder`

phated: `configFilesBase` can just be `configFiles`

lib/versioned/^3.7.0/index.js
phated: Put the `isString` check back in when the validation is removed.

lib/versioned/^4.0.0-alpha.1/index.js
```diff
@@ -44,8 +43,8 @@ function execute(opts, env) {
     }
     if (opts.tasks) {
       var tree = {};
-      if (opts.description && isString(opts.description)) {
-        tree.label = opts.description;
+      if (config.description) {
```
phated: Put the `isString` check back in when the validation is removed.

lib/versioned/^4.0.0-alpha.2/index.js
```diff
@@ -49,8 +48,8 @@ function execute(opts, env) {
     }
     if (opts.tasks) {
       tree = gulpInst.tree({ deep: true });
-      if (opts.description && isString(opts.description)) {
-        tree.label = opts.description;
+      if (config.description) {
```
phated: Put the `isString` check back in when the validation is removed.

lib/versioned/^4.0.0/index.js
```diff
@@ -49,8 +48,8 @@ function execute(opts, env) {
     }
     if (opts.tasks) {
       tree = gulpInst.tree({ deep: true });
-      if (opts.description && isString(opts.description)) {
-        tree.label = opts.description;
+      if (config.description) {
```
phated: Put the `isString` check back in when the validation is removed.

lib/versioned/^4.0.0/index.js
```diff
@@ -59,8 +58,8 @@ function execute(opts, env) {
     }
     if (opts.tasksJson) {
       tree = gulpInst.tree({ deep: true });
-      if (opts.description && isString(opts.description)) {
-        tree.label = opts.description;
+      if (config.description) {
```
phated: Put the `isString` check back in when the validation is removed.

lib/versioned/^4.0.0-alpha.2/index.js
```diff
@@ -59,8 +58,8 @@ function execute(opts, env) {
     }
     if (opts.tasksJson) {
       tree = gulpInst.tree({ deep: true });
-      if (opts.description && isString(opts.description)) {
-        tree.label = opts.description;
+      if (config.description) {
```
phated: Put the `isString` check back in when the validation is removed.

package.json
```diff
@@ -26,22 +26,23 @@
...
     "pretest": "npm run lint",
-    "test": "mocha --async-only --timeout 3000",
+    "test": "mocha --async-only --timeout 3000 test/lib test",
     "cover": "nyc --reporter=lcov --reporter=text-summary npm test",
     "coveralls": "nyc --reporter=text-lcov npm test | coveralls",
     "changelog": "github-changes -o gulpjs -r gulp-cli -b master -f ./CHANGELOG.md --order-semver --use-commit-body"
   },
   "dependencies": {
     "archy": "^1.0.0",
+    "camelcase": "^3.0.0",
```
phated: This can be removed when `excludeCliArgs` is removed.

index.js
phated: `toConsole` can be updated to remove previous listeners if called a 2nd time.  You can then use it at the beginning of this file and call it again after you normalize the config.

sttk:
@phated Thanks for your reviewing. I'll address what you pointed out this week end.

lib/shared/config/convert-config.js
sttk: I got it. I'll remove these validations.

But only one thing, converting the path of `flags.gulpfile` is necessary because users may set a relative path from the config file.

phated: Can that be done inline instead of an abstraction like this (similar to configBase)?

sttk: Oh, it is possible to do so. Thanks!

sttk:
@phated I've modified.

lib/shared/config/load-files.js
```diff
@@ -0,0 +1,26 @@
...
+
+  configFileOrder.forEach(function(key) {
+    var filePath = configFiles[key];
+    if (!filePath) {
+      return;
+    }
+
+    copyProps(require(filePath), config, function(value, name) {
+      if (name === 'flags.gulpfile') {
+        return path.resolve(path.dirname(filePath), value);
```
phated: Oh interesting, this is making the path in the config file relative its location.

lib/shared/config/load-files.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+var copyProps = require('copy-props');
+var path = require('path');
+
+function loadConfigFiles(configFiles, configFileOrder) {
+  var config = {};
+
+  configFileOrder.forEach(function(key) {
```
phated: Named function (I can make this change during merge).

lib/shared/config/load-files.js
```diff
@@ -0,0 +1,26 @@
...
+    var filePath = configFiles[key];
+    if (!filePath) {
+      return;
+    }
+
+    copyProps(require(filePath), config, function(value, name) {
```
phated: Named function (I can make this change during merge).

lib/shared/log/toConsole.js
```diff
@@ -14,6 +14,9 @@ var levels = [
 ];
 
 function toConsole(log, opts) {
+  // Remove previous listeners to enable to call this twice.
+  log.removeAllListeners();
```
phated: I think this should be more specific to avoid removing other listeners someone else might have attached (we are using a global, after all).  I can change this in the merge.

phated:
@sttk amazing work, as always.  I'll get this merged and make a couple changes I commented about.

lib/shared/log/toConsole.js
sttk: Oh, you are exactly right. Thanks.

sttk:
@phated Thank you for merging!

PR #101, opened 2017-01-11
added node v7 to travis and appveyor

JPeer264:
Node v7 should be included into the test suites. Many people are using v7 already.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: false
 language: node_js
 node_js:
+  - "7"
   - "6"
   - "5"
   - "4"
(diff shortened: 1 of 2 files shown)


phated:
People really shouldn't be using v7 in real environments.  Also, we aren't able to test old gulp versions on v7 so I don't think we should include it.  I was told not to worry about it until v8

JPeer264:
That's a good point. Ok for me it's new to not worry about it until v8, but I guess there are many good reasons

phated:
I've been told odd versions are experimental and even versions are the release.  After an even version is cut, the odd version is no longer supported.  We should probably remove the odd releases from our test matrix.

JPeer264:
That makes sense though. I am not sure about removing it. As long as they do not make noise and fail the tests it would be acceptable.

JPeer264:
@phated I removed now all uneven node versions from `.travis.yml` and `appveyor.yml`

phated:
Thanks @JPeer264

PR #105, opened 2017-02-24
Docs: Clarify CLI semantics when listing more than one task

kylecordes:
Make the same CLI doc change here as in the gulpjs repo:

https://github.com/gulpjs/gulp/pull/1916

PR #106, opened 2017-02-27
Make README match CLI.md multiple task explanation

kylecordes:
... as requested.

https://github.com/gulpjs/gulp-cli/pull/105#issuecomment-282493599


phated:
Many thanks @kylecordes

PR #107, opened 2017-04-25
Update: Add flags.continue to config

phated:
@sttk I wanted to see if I could understand the process of adding a config property so I attempted to add the `--continue` flag.  Can you review?

### DIFF
--- a/index.js
+++ b/index.js
@@ -59,14 +59,6 @@ var usage =
 var parser = yargs.usage(usage, cliOptions);
 var opts = parser.argv;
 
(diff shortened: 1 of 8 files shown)


sttk:
@phated I've reviewed your code, and your work is excellent!

About the process of adding a config property, I think it's not necessary to add a test of `test/lib/config-cli-flags.js` but it is also good, of course.

I've merged this PR.

phated:
Thanks @sttk!  Glad you noticed that I added you to the gulp-cli team.  Welcome!

Issue #110, opened 2017-04-26
Expose `flags.logLevel` as config option


phated:
This option is very important to a lot of people (I think).

PR #113, opened 2017-05-27
Add config `flags.logLevel`

sttk:
I implemented for the issue #110.

Since Appveyor failed, I modified `appveyor.yml`, too.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,6 +10,8 @@ environment:
     - nodejs_version: "6"
 
 install:
+  - npm -g install npm@2
+  - set PATH=%APPDATA%\npm;%PATH%
   - ps: Install-Product node $env:nodejs_version
   - npm install
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,11 +1,6 @@
 # http://www.appveyor.com/docs/appveyor-yml
 # http://www.appveyor.com/docs/lang/nodejs-iojs
 
-branches:
-  # whitelist
-  only:
-    - master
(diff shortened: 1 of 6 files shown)


sttk:
I've modified according to #114

appveyor.yml
```diff
@@ -15,8 +10,6 @@ environment:
     - nodejs_version: "6"
 
 install:
-  - npm -g install npm@latest
```
phated: Are we going to run into weird test runs on old node versions with this removed?

sttk: @phated The reason of removing this line is that the latest npm don’t support old versions of nodejs (< v4.x). This is needed if we are going to support gulp-cli on these old versions and use Appveyor for test on Windows.

Or we might be able to skip test on Windows about old versions of nodejs.

phated: Not dropping 0.10. 

Just install npm at a good version, like 2

sttk: @phated If the “weird test” means using gulp-test-tool, since this tool is a thing to solve [this issue](https://www.appveyor.com/docs/lang/nodejs-iojs/#garbled-or-missing-output) on nodejs v0.10 on Windows, we can remove this tool if we skip test on Windows about old versions of nodejs.

sttk: My reply crossed with yours.

> Just install npm at a good version, like 2

I got it. I checked our test succeeded with `npm@2` now.

sttk:
@phated Thanks for merging.

PR #118, opened 2017-08-04
Fix: Blacklist redirect

erikkemperman:
Several tests were failing for me [here](https://travis-ci.org/erikkemperman/gulp-cli/builds/260972095?utm_source=github_status&utm_medium=notification), while the exact same code on [gulp/master](https://travis-ci.org/gulpjs/gulp-cli/builds/242142880?utm_source=github_status&utm_medium=notification) was somehow fine.

Superficially the errors seemed to be about flags, such as loglevel... But it was something else entirely.

Turns out an external change sabotaged a bunch of our tests here, viz. we now get a "moved permanently"  (301) when fetching the blacklist, and apparently the wreck library doesn't follow redirects (by default).

I fixed the URL in this PR, perhaps the wreck options should be revisited some time.

This was pretty annoying to track down... The tests are already slightly tricky to debug because stdout and stderr are hijacked. But it becomes really problematic if a unit test depends on unrelated components, both internal (in this case, the getBlacklist module) and external (the corresponding webserver).

IMHO, it should not be possible for this kind of external change to break unit tests?

### DIFF
--- a/lib/shared/getBlacklist.js
+++ b/lib/shared/getBlacklist.js
@@ -2,7 +2,7 @@
 
 var wreck = require('wreck');
 
-var url = 'http://gulpjs.com/plugins/blackList.json';
+var url = 'https://gulpjs.com/plugins/blackList.json';
 
 function getBlacklist(cb) {
   wreck.get(url, { json: true }, function(err, res, blacklist) {


sttk:
@erikkemperman I'm sorry my modification and `gulp-test-tool` annoyed you.

About the unit test of logLevel, using `--verify` flag is the only way for testing `log.warn`.

About `gulp-test-tool`, this tool is a thing to solve [this issue](https://www.appveyor.com/docs/lang/nodejs-iojs/#garbled-or-missing-output) on nodejs v0.10 on Windows, though I'm worried someone would say that.

I wish there are better alternatives...

erikkemperman:
Maybe "annoying" was the wrong word, let's make it "complicated". I wanted to submit a really simple PR but got stuck on all these failing tests :-)

If you agree this solves things at least for the time being, could you review and merge? Thanks!

sttk:
@erikkemperman I've reviewed this PR and it's good.
I didn't notice that the url of gulp's plugins registory was changed as http -> https 9 days ago.
Thank you for fixing this issue.

erikkemperman:
@sttk Thanks for review and merge! Unfortunately my day is almost over, so my simple little PR will have to wait a bit.

phated:
@erikkemperman sorry this tripped you up.  Testing CLI tools is quite tricky in node, but hopefully we can make incremental improvements over time.

Also note that the SSL change was sprung on me and I didn't have a chance find everything that would be affected.  Thanks for fixing this; I'll get it shipped in a patch.

phated:
Looks like we have some features and updates queued up so this will go out as 1.4.0

PR #121, opened 2017-08-08
Update: -f alias for --gulpfile

erikkemperman:
I meant to propose this for some time -- I find myself using `--gulpfile` quite often, and would love to be able to use an alias for that flag.

Not sure if there's a policy on this, but `-f` makes sense to me and would mesh nicely with what folks might know from other build tools.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ __Some flags only work with gulp 4 and will be ignored when invoked against gulp
     </tr>
     <tr>
       <td>--gulpfile [path]</td>
-      <td></td>
+      <td>-f</td>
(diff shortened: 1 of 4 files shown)


phated:
`-f` for "file"?  I like keeping this aligned with other tools (which ones do you mean?)

erikkemperman:
Yeah, that works :-)

Make, Rake, Jake, Ant and Maven all use `-f`, probably others...

phated:
I think any additional flags we add need to be a breaking version because someone might be using it internally to their gulpfile.

phated:
Thanks @erikkemperman - I'm bundling a bunch of changes for the 2.0 push here.

erikkemperman:
Ah, forgot about this one. Thanks!

Issue #93, opened 2016-08-24
Config option to multiple command-line tasks in series


erikkemperman:
It isn't really explicitly documented that commandline tasks will be run in parallel, unless I missed something... But I guess users of almost any other build tool would expect them to be done in series. E.g., `gulp clean build` seems like a common scenario.

So I was wondering:
- should the default be to run commandline tasks in a `series` call, instead?
- if not, should there maybe be an option to run tasks in `series`?
- either way, should this be documented more explicitly?

Thanks, I hope it's not a silly suggestion...

Ref. https://github.com/gulpjs/gulp-cli/blob/master/lib/versioned/%5E4.0.0/index.js#L77

phated:
The concept of gulp has always been to run everything at "maximum concurrency" and I think that is the proper behavior to propagate to the CLI.  The recommended way to run gulp is with a single task (or, more likely, none for `default`), even though we support multiple.  I agree that this could be documented better (like __many__ things in gulp).

I'll leave this open incase others have opinions.

contra:
Yeah, we should document this better. The right way to do this would be a new task that runs those two in series, but I could see how this is confusing.

That said, I don't think a flag would be terrible here to reduce boilerplate.

phated:
We are kinda going overboard with flags.  Maybe this could be configurable in the new `.gulp.*` configuration files.

erikkemperman:
Yeah, coming from gulp3 I guess it won't be unexpected, but now that there's undertaker in place of orchestrator people have to compose with series/parallel explicitly anyway, and I thought it might make sense to extend that to the top (cli) level. 

Coming from any other build tool it might cause some confusion initially.

Of course ai can just make special tasks -- actually looking at some clutter in my own gulpfiles (tasks like `cleanbuild`) made me wonder about this.

Or I could always have bash handle this (`gulp clean && gulp build`). But maybe an option or config to change the cli composition might make some sense.

phated:
@erikkemperman I typically defer "conditional" tasks (like clean) to flags and then have code that generates the sequence inside the gulpfile.

```bash
> gulp build --clean
```

```js
gulp.task('build', (argv.clean ? gulp.series(clean, build) : build));
```
Or something similar.

Deferring to bash is also a good idea but the CLI startup time might be really slow (not sure).

erikkemperman:
Ah, that's a nice pattern. No wonder you're hesitant to have the cli claim options 👍 

[...]

PR #122, opened 2017-08-08
Update: Add flag/config for running tasks from CLI in series

erikkemperman:
https://github.com/gulpjs/gulp-cli/issues/93

Sorry for all the PRs, going away for a couple of days and wanted to get this out.

I figured this ought to be a flag too, after all, since I imagine it might not usually make sense to do this per-module, but rather for occasional single runs of the CLI.


phated:
@erikkemperman I'm more inclined to take this as a "config only" option.  I'm hesitant to add more CLI options (people already complain we collide with their custom options).

erikkemperman:
@phated

Well, okay, easy enough to take the flag stuff out of the PR of course. But I hope you don't mind one little attempt at persuasion, because I for one would probably want to use this for occasional runs and perhaps not always as a structural element of my build...

I get the complaints about colliding with people's custom flags, but might argue the one exception would be if gulp-cli now implemented the same functionality out of the box that I had to do myself before.

And with "series" it seems likely that if folks built that themselves, it would probably be doing something very much like what that flag would do in this PR?

But anyway, having said that, if you're adamant about it I'll happily amend the PR, probably some time this week.

phated:
@sttk do you have any thoughts about this? Specifically whether it should be config-only or config + flag

sttk:
@phated I think both cli flag and config are good, and this flag will be useful.
Adding a new flag is worried about conflicts with user-defined flags, but there would be few cases about this flag.

phated:
Cherry-picked this into master as c454a6d2196c02b7b8cb891b73b357d43fb03e8e and did some conflict resolution/cleanup.

@erikkemperman Thanks for your patience with me on this - I'm going to land it as a "new" feature in 2.0 incase it is breaking for anyone relying on `--series` within their gulpfile.

erikkemperman:
@phated That’s great, thanks a lot!

PR #130, opened 2017-08-19
Modify the tests about log-level config against timeout on AppVeyor

sttk:
The tests about log-level config and verify flag cause timeout errors frequently on Appveyor, because accessing to the remote blacklist file sometimes take too much time.

So I've modified the test cases about log-level config to replace/skip accessing the blacklist file. (Replacing all cases is impossible because `log.warn` is used by only verify flag in gulp-cli).

### DIFF
--- a/test/config-flags-log-level.js
+++ b/test/config-flags-log-level.js
@@ -2,8 +2,12 @@
 
 var expect = require('expect');
 var path = require('path');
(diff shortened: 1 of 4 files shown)


phated:
@sttk is it still happening in this?

sttk:
@phated It is seemed that this happened temporarily. I'll close this pr.

PR #132, opened 2017-08-20
Update yargs to 7.1.0

### LINKED ISSUE #127: Update yargs
The newest versions stopped supporting node <4 but it's still the best argument parser so we should stick with it but we should update to the last version they released with node 0.10 support.

tdeschryver:
Closes #127

### DIFF
--- a/index.js
+++ b/index.js
@@ -107,7 +107,7 @@ function handleArguments(env) {
   toConsole(log, opts);
 
   if (opts.help) {
-    console.log(parser.help());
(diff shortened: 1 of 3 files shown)


phated:
Just going to group this with breaking because I don't want to sort through their changelog.

phated:
Merged as 2e87ac8aefb6b4228deb76dcbbec8b8763249330 - had to manually rebase it.

phated:
Thanks @tdeschryver!

PR #133, opened 2017-08-21
Add update notifier

### LINKED ISSUE #65: Add update notifier
Is this something people like and would want?  We would use https://github.com/yeoman/update-notifier like yeoman does.

cc @sindresorhus

tdeschryver:
Closes #65

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "pretty-hrtime": "^1.0.0",
     "semver-greatest-satisfied-range": "^1.0.0",
     "tildify": "^1.0.0",
+    "update-notifier": "^1.0.3",
     "v8flags": "^2.0.9",
     "wreck": "^6.3.0",
     "yargs": "^3.28.0"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -47,6 +47,7 @@
     "pretty-hrtime": "^1.0.0",
     "semver-greatest-satisfied-range": "^1.0.0",
     "tildify": "^1.0.0",
+    "update-notifier": "^1.0.3",
```
tdeschryver: This is the latest version supporting node 0.10.

phated:
Thanks @tdeschryver - we're going to hold off on this a little bit because it adds a lot of overhead to startup time and I want to use libraries that actually support node 0.10 at latest.

tdeschryver:
@phated sure, no problem. 
I was a bit hesitant for the same reason you mentioned.
I'll take a look at the open issues in the next week if that's OK :smile:

phated:
I'm going to close this for the reasons above.  We can revisit the issue down the road.

Issue #63, opened 2016-02-28
Allow customizing CLI colors through .gulprc file


sttk:
I've added customizing mechanism in gulp-cli.
And also I've made colors of a task tree customizable with this.

[sttk/gulp-cli#make_customizable](https://github.com/gulpjs/gulp-cli/compare/master...sttk:make_customizable)
- You can change the colors by `.gulprc` files in user home directory or project working directory.
- You can also change the colors with a file specified by `--customize` option.
- You can see a list of customizable settings by `--customize` option which has no value.

phated:
@sttk The first thing that I'd like to address is the `.gulprc` file.  I am all for having it, but I think we need it implemented over in Liftoff (which has been discussed before) - Check out the issue at https://github.com/js-cli/js-liftoff/issues/8 and maybe you'd be interested in implementing the rc lookup over there before submitting this PR to gulp-cli.

sttk:
@phated OK. I'll see the issue and Liftoff. And I'll try it if I can.

phated:
Changed the title here.  Also split the .gulprc file to issue https://github.com/gulpjs/gulp-cli/issues/69

phated:
Random ideas for theming (so they are documented somewhere):
```js
// regexp? HIGHLIGHT: (.*)

log.withTheme()

log('Write some text and {HIGHLIGHT: The Thing}', theme)

log('{ERROR: Failed to load external module} {HLIGHT: {NAME}}', {NAME: name})

log('{ERROR}Failed to load external module{/ERROR} {HIGHLIGHT}' + name + '{/HIGHLIGHT}')

log.info('Using gulpfile {HIGHLIGHT: %s}', tildify(env.configPath));
```

sttk:
@phated I've created a package [theming-log](https://github.com/sttk/theming-log) to manage text decoration of logging with themes on the trial basis.

This package is still a prototype and would be needed to add some functions (for example, printf-style variables).

Can you have a look whether it fits your idea?

sttk:
@phated Thank you for having a look, and I'm glad you like it.

About the way to using **theming-log** for this package, most of logs in `gulp-cli/index.js`, including the help message, would be ok in the normal way because they are outputted after loading the config files, but for adding themes to the help message it may be needed to overide some methods of **yargs**.  

Since logs for `require` and `requireFail` events are called before loading the config files, they would be needed to think out.

[...]

PR #134, opened 2017-08-22
Replace chalk

### LINKED ISSUE #125: Replace chalk
I'm thinking we should replace chalk with individual `ansi-*` modules from @jonschlinkert: https://github.com/jonschlinkert?utf8=%E2%9C%93&tab=repositories&q=ansi&type=&language=

tdeschryver:
Closes #125


test/shared/stripAnsi.js
```diff
@@ -0,0 +1,11 @@
+var ansiPattern = [
+  '[\\u001B\\u009B][[\\]()#;?]*(?:(?:(?:[a-zA-Z\\d]*(?:;[a-zA-Z\\d]*)*)?\\u0007)',
+  '(?:(?:\\d{1,4}(?:;\\d{0,4})*)?[\\dA-PRZcf-ntqry=><~]))',
+].join('|');
+var ansiRegex = new RegExp(ansiPattern, 'g');
+
+function stripAnsi(value) {
```
tdeschryver: Maybe it would be better to move this to `gulp-test-tools` if possible?

phated:
I like where this is going but maybe we should combine it with the work on #63 - @sttk what do you think?

sttk:
@phated Please merge this first. I'll address #63 later.

test/shared/stripAnsi.js
sttk: @tdeschryver I also think it's good and welcome to send a pr to `gulp-test-tools`.

sttk:
Does `ansi-*` support `--no-color` flag?

tdeschryver:
@sttk it does has a [reset](https://github.com/jonschlinkert/ansi-reset) and it also has a [wrap](https://github.com/jonschlinkert/ansi-wrap). 
If this is what you're looking for :smile:

test/shared/stripAnsi.js
tdeschryver: OK, I'll try to open a PR later today!

tdeschryver: Or even better, I think we should use [ansi-strip-color](https://github.com/jonschlinkert/strip-color)

sttk:
@tdeschryver `chalk` has a feature to disable coloring by `--no-color` flag, and `gulp-cli` publishes this flag as its command line option. So we should support this flag even when using `ansi-*`.

phated:
Looks like this still needs to support `--no-color` and there are some files that got changed (fixture `package.json` files)

tdeschryver:
Oops, totally forgot about this 😅, I'll add during this week!

phated:
Thanks! I'm guessing you won't need the `strip-ansi` stuff in the tests once that's supported

tdeschryver:
Hmmm weird, seems like those fixtures which were added got changed from `CRLF` to `LF`? :confused: 
Anyways, I reverted the fixture changes.

phated:
@tdeschryver https://github.com/doowb/ansi-colors now inlines all the colors.  This will likely be the best solution for us.  I can look into making those changes if you don't have time.

tdeschryver:
@phated I'm currently a bit busy, if you need it right now it would be best if someone else takes this.
If you can wait a week, 2 at max, I can take it.

phated:
Closed by #138

PR #135, opened 2017-09-24
Use meaningful processTitle for better monitoring

nguymin4:
At the moment when we run several gulp tasks separately, they have the same process title `gulp` which is very hard to differentiate and monitor them. This PR is to solve that issue.

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,8 +33,10 @@ var ranges = fs.readdirSync(__dirname + '/lib/versioned/');
 // before anything touches it
 process.env.INIT_CWD = process.cwd();
 
+var cmdArgs = process.argv.slice(2);
 var cli = new Liftoff({
   name: 'gulp',
+  processTitle: ['gulp'].concat(cmdArgs).join(' '),
   completions: completion,
   extensions: interpret.jsVariants,
   v8flags: v8flags,


phated:
Thanks for submitting this here.

index.js
```diff
@@ -33,8 +33,10 @@ var ranges = fs.readdirSync(__dirname + '/lib/versioned/');
 // before anything touches it
 process.env.INIT_CWD = process.cwd();
 
+var cmdArgs = process.argv.slice(2);
 var cli = new Liftoff({
   name: 'gulp',
+  processTitle: ['gulp'].concat(cmdArgs).join(' '),
```
phated: I think we can make the process title nicer.  Maybe we should only show the tasks being run?  Maybe we should wrap them in `(  )`

nguymin4: I can wrap them in  `( )` but I think we should keep the original command argv without any modification. So for example `gulp watch` should be different from `gulp watch -L`. Is that ok for you?

phated: I'm a bit concerned that the process title will get too long.  Maybe it should get truncated?

phated: @sttk any ideas about how we can reduce the size of the process title but still show uniqueness across flags?

sttk: @phated I have no idea to make short it.

I agree with @nguymin4's saying, keeping the original command argv, but there may be some platforms or tools which are not support too long process name. (Though I don't know actual cases.)

 I think it would be better that this feature is optional and able to be specified in config file.

nguymin4: Actually, I cannot come up with a case which can have so long command. If it's too long, then you are doing something wrong, aren't you?

sttk: @nguymin4 My thought seemed to be overcautious. I think no problem about this now.

phated: Alright, if @sttk doesn't see a problem with it.  Can you turn this into a named utility function?

phated:
Rebased as e5fdefb29e65c441ada61ba8308adec4f045c53f - Thanks @nguymin4

PR #141, opened 2018-01-01
Update devDependencies.

sttk:
I've updated devDependencies in package.json and have made tests succeed.


package.json
```diff
@@ -24,7 +24,8 @@
   ],
   "scripts": {
     "lint": "eslint . && jscs index.js bin/ lib/ test/",
-    "prepublish": "marked-man --name gulp docs/CLI.md > gulp.1",
+    "prepare": "marked-man --name gulp docs/CLI.md > gulp.1",
+    "prepublishOnly": "npm run prepare",
```
phated: I still use npm@2 so this won't work for me.

.gitignore
```diff
@@ -14,3 +14,4 @@ test/out-fixtures/*
 test/output/
 test/watch-*.txt
 gulp.1
+package-lock.json
```
phated: Makes sense but does npm pick up `.npmrc` files from the repo?  Might be better to include an npmrc that stops the package-lock creation.  Also, belongs in a separate commit.

package.json
```diff
@@ -51,20 +52,20 @@
...
+    "babel-preset-es2015": "^6.24.1",
     "babel-register": "^6.5.1",
-    "coveralls": "^2.7.0",
+    "coveralls": "^2.13.1",
     "eslint": "^1.7.3",
-    "eslint-config-gulp": "^2.0.0",
+    "eslint-config-gulp": "^2.1.0",
     "expect": "^1.20.2",
-    "gulp": "gulpjs/gulp#4.0",
+    "gulp": "^4.0.0",
```
phated: I think we should only update this in the dep update PR.

test/fixtures/packages/invalid-package.json
```diff
@@ -1,21 +1,21 @@
...
-    "node": ">= 0.9"
-  },
-  "licenses": [
-    {
-      "type": "MIT",
-      "url": "https://raw.githubusercontent.com/gulpjs/gulp/master/LICENSE"
-    }
-  ]
-}
+{
```
phated: These files were causing problems on my machine due to weird line ending stuff.  I had to tell git to ignore the changes to make them go away.

package.json
sttk: Sorry. `prepare` and `prepublishOnly` was introduced in npm@4. I'll put back this.

.gitignore
sttk: I got it. I'll commit .npmrc separately.

sttk:
@phated I've removed unnecessary modifications and have separated into commits of updating dev dependencies and disabling package-lock.json.

package.json
```diff
@@ -57,13 +57,13 @@
     "eslint": "^1.7.3",
     "eslint-config-gulp": "^2.0.0",
     "expect": "^1.20.2",
-    "gulp": "gulpjs/gulp#4.0",
+    "gulp": "^4.0.0",
     "gulp-test-tools": "^0.6.1",
-    "jscs": "^2.3.5",
+    "jscs": "^3.0.7",
```
phated: I thought jscs changed a bunch of stuff in 3.0 that made our config file invalid?

package.json
```diff
@@ -57,13 +57,13 @@
...
     "jscs-preset-gulp": "^1.0.0",
-    "marked-man": "^0.1.3",
+    "marked-man": "^0.2.1",
```
phated: What changed in this module?

package.json
sttk: [Release note of jscs 3.0](https://github.com/jscs-dev/node-jscs/releases/tag/v3.0.0) wrote:
> We didn't rename/remove any rules. ...

So I think there are no problem to use 3.0.

package.json
sttk: Just I think newer module is better. Doesn't `^0.1.3` represents the range `>=0.1.3 <0.2.0`?

phated:
@sttk can you also bump minor versions for any dependencies that are above 1.0?  Ref #147

sttk:
I've update more dependency versions,  fancy-log and so on.

phated:
@sttk I cherry-picked your commits into master and squashed some stuff (needed to deal with line endings of those fixtures on my new computer) - I'll get this published with the exitCode thing.

sttk:
@phated I got it. Thanks!

Issue #139, opened 2017-12-17
Include 3.x in test matrix?


phated:
I was just looking at #115 and realized that we don't have any coverage/testing over the 3.x CLI usage.  It'd be nice to add that to the CI matrix for more peace-of-mind.

PR #145, opened 2018-01-13
Return exit code 1 when incomplete tasks are detected

barbogast:
Looks like this could be a fix for #2081

### DIFF
--- a/lib/versioned/^4.0.0/log/sync-task.js
+++ b/lib/versioned/^4.0.0/log/sync-task.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var log = require('gulplog');
+var process = require('process');
 var ansi = require('../../../shared/ansi');
 
 var tasks = {};
@@ -15,6 +16,8 @@ function warn() {
   var taskNames = taskKeys.map(function(key) {
     return tasks[key];
   }).join(', ');
+
+  process.exitCode = 1;
 
   log.warn(
     ansi.red('The following tasks did not complete:'),


phated:
Please add tests and make sure you aren't breaking old versions.   `process` isn't a module and is just a global.

phated:
Sure! Anything that is easily backported should be.

sttk:
@barbogast Could you also implement for gulp v3? Regarding tests on v3, since we will address them in the issue #139, you may write only a unit test for [`logEvents`](https://github.com/gulpjs/gulp-cli/blob/master/lib/versioned/%5E3.7.0/log/events.js#L11) in `events.js`.

barbogast:
I just added a test and removed the require('process') call.

I'd try to backport the change to gulp v3 but to me it looks like gulp v3 doesn't detect incomplete tasks but just exits the process.

The integration tests fail for nodejs 0.10 as expected (see https://github.com/gulpjs/gulp/issues/2081#issuecomment-355982035). Is there a way to skip the test for nodejs 0.10?

phated:
```js
if (!('exitCode' in process)) {
  this.skip();
}
```
Should work???

barbogast:
I'm pretty sure that this wouldn't solve the problem. The code didn't crash. Only the test failed because the process didn't exit with an error code (as is expected for nodejs 0.10). We need a way to disable this test case for nodejs 0.10. We could parse `process.version` and skip the test for nodejs < 0.11. Kind of hacky, bit it should work...

phated:
That goes in the test to skip it. The property won't exist in old node

barbogast:
Oh, sorry, I didn't understand. I just checked, and `exitCode` is not present on `process` (except when it's set explicitly).

We could detect the node version using https://github.com/phated/sver-compat

test/lib/sync-task.js
```diff
@@ -0,0 +1,43 @@
...
+
+function nodeDoesNotNodeSupportExitCode() {
+  var currentVersion = new Semver(process.versions.node);
+  var minimalVersion = new Semver('0.11.0');
+  return currentVersion.lt(minimalVersion);
+}
+
+describe('sync-task', function() {
+  it('should return error code 1 if any tasks did not complete', function(done) {
+    var this_ = this;
```
sttk: @barbogast It's not necessary to use the new package. As @phated said, this test case can be skipped by the following code at this point:

```js
  it('should return error code 1 if any tasks did not complete', function(done) {
    if (!('exitCode' in process)) { // On v0.10 only, this condition is true because process doesn't have .exitCode.
      this.skip();
      return;  // This is not necessary because skip() throws an exception, but is recommened to avoid confusing.
    }
    ...
```

barbogast: Are you sure this will work? I just rechecked and 

```
> 'exitCode' in process
false
```
for 
```
node -v
v8.9.3
```

Or am I misunderstanding completely?

barbogast: Ah, I think I understood. Do you mean that `exitCode` was set already by the new line in `sync-task.js`? This would work, but only if `require('gulp-test-tools').gulpRunner` which is used by the test doesn't spawn a new process, right? I guess that it does but I'll try when I get home.

sttk: @barbogast I'm sorry, I really misunderstood. You are right. I did only checked process.exitCode is undefined only on v0.10 and I completely thought process.exitCode is zero on other versions.

sttk: @barbogast Since these test cases in `test/lib/sync-task.js` are not unit test, please move this file to `test/` and rename it to another like `exit-code.js`.

In addition, I think it's better that these two cases can be put into one case and then `nodeDoesNotNodeSupportExitCode()` is used not for `this.skip()` but for unchecking exit code.

package.json
```diff
@@ -64,7 +64,8 @@
     "marked-man": "^0.1.3",
     "mocha": "^3.2.0",
     "nyc": "^10.0.0",
-    "rimraf": "^2.6.1"
+    "rimraf": "^2.6.1",
+    "sver-compat": "^1.5.0"
```
phated: Don't bring in this dep.

test/lib/sync-task.js
```diff
@@ -0,0 +1,43 @@
...
+var runner = require('gulp-test-tools').gulpRunner;
+
+
+function nodeDoesNotNodeSupportExitCode() {
+  var currentVersion = new Semver(process.versions.node);
+  var minimalVersion = new Semver('0.11.0');
+  return currentVersion.lt(minimalVersion);
+}
+
+describe('sync-task', function() {
```
phated: Use this instead:

```js
beforeEach(function(cb) {
  if (process.version.slice(5) === 'v0.10') {
    this.skip();
  }

  cb();
});

File: test/lib/sync-task.js (hunk 1 above)
Comment (phated) [reply test/lib/sync-task.js]: Please move the file like sttk mentioned and remove that utility (using my other recommendation above).

File: test/fixtures/gulpfiles/gulpfile.js [hunk 4]
```diff
@@ -20,6 +20,15 @@ function errorFunction() {
 function anon(cb) {
   cb();
 }
+
+function notCompleting1() {
```
phated: Do we need 2 non-completing tasks?

barbogast: I created 2 tasks to also test that the log message is rendered correctly with multiple non-completing tasks. Should I remove the second task?

phated: That's fine.  I'm not partial either way.

barbogast:
Thank you for your review. :+1: 

The new dependency and the utility function are removed.

Note that in addition to a test for the exit code I added a test for the log message about the non-completing tests. I couldn't find any tests for `sync-tasks.js` so I went ahead and added a second test case.

The reason I created 2 test cases is that test case for the log message can run under nodejs 0.10 and should be reported as successful while the second test case is then reported as skipped. Should I join them anyway?

The test file is moved to the `test`-directory. I have not yet renamed the test file since it not only tests the exit code but also the log message for non-completing tests. So I think the name `exit-code.js` is not fitting. Maybe `non-completing-tasks.js`?

phated:
non-completing-tasks.js sounds great!

barbogast:
all right, I'll rename it

phated:
This PR looks great @barbogast - I'm just going to wait until the CI complete then merge and publish

barbogast:
Great, thanks a lot for your reviews, @phated @sttk  

Should I sqash the commits before merging?

phated:
> Should I squash the commits before merging?

Nah, I'll do that in the merge.

sttk:
@barbogast Great! Thank you for your contribution.

phated:
Published as 2.0.1 - Thanks again @barbogast!!

PR #147, opened 2018-01-17
Fix: Update fancy-log dependency, gulpjs/gulp#2098

kayo5994:
Fix for gulpjs/gulp#2098.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "concat-stream": "^1.6.0",
     "color-support": "^1.1.3",
     "copy-props": "^2.0.1",
-    "fancy-log": "^1.1.0",
+    "fancy-log": "^1.3.2",
     "gulplog": "^1.0.0",
     "interpret": "^1.0.0",
     "isobject": "^3.0.1",


phated:
The semver range we've specified in the package.json already accounts for the version range you are trying change.  It's likely that your problem is a `package-lock.json` file and not the way we specify dependency ranges.

demurgos:
Shouldn't the semver range be updated anyway?
If this library does not work properly unless the version of `fancy-log` is at least `1.3.0` then it should state so in the package file.

I also think that in a proper environment and a fresh install there's no big risk but you can always find a way for it to break.
Assume that I have a dependency on a plugin with an exact dependency on `fancy-log@1.2.0`, with npm's installation algorithm it may install this version first and then when it reaches `gulp-cli` it will not install the latest version but reuse the older one (and exhibit the issue, even without a lock file).

phated:
> Assume that I have a dependency on a plugin with an exact dependency on fancy-log@1.2.0, with npm's installation algorithm it may install this version first and then when it reaches gulp-cli it will not install the latest version but reuse the older one 

Yeah, I guess npm does that now with their automatic deduping algorithm.  This is why I should have never upgrade from node 0.10/npm 2 😛 

I'll have @sttk bump the minor versions for deps in https://github.com/gulpjs/gulp-cli/pull/141 since it all falls under "Upgrade".

phated:
Published as 2.0.1

PR #149, opened 2018-01-20
Return exit code 1 when incomplete tasks are detected on gulp v3

sttk:
I've implement for gulp v3 to return exit code 1 and print warning messages when there are incomlete tasks. (Related: #145 and [gulpjs/gulp#2081](https://github.com/gulpjs/gulp/issues/2081).)

### DIFF
--- a/lib/shared/exit.js
+++ b/lib/shared/exit.js
@@ -2,6 +2,7 @@
 
 // Fix stdout truncation on windows
 function exit(code) {
+  /* istanbul ignore next */
   if (process.platform === 'win32' && process.stdout.bufferSize) {
     process.stdout.once('drain', function() {
       process.exit(code);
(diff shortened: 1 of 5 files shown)


phated:
I don't believe we can successfully detect incomplete tasks because gulp 3 supported sync tasks. So everything "completed"

sttk:
@phated gulp receives the events of starting and finishing a task, and outputs these logs. So incomplete tasks can detect by finding tasks which started but not finished at the end of gulp.
This pr implements this process simply. This works well about sync tasks, too.

BTW, I'm sorry but I found a bug in this pr. If a task error occured, this pr outputs incomplete tasks warning though gulp4 does not output.

sttk:
@phated I wrote the wrong thing. As you wrote, gulp v3 cannot detect incomplete task which is a sync task but execute async process in it. Umm... Should we cancel this pr?

phated:
@sttk yeah, you found what I was talking about.  I'll close this as it doesn't make sense for 3.x

PR #151, opened 2018-01-23
Fix: Log requireFail results in yellow on stdout instead of red on stderr (fixes #142)

### LINKED ISSUE #142: Revisit requireFail color and message
It seems that some consumers are allergic to the color red - we should revisit the coloring and messaging of the `requireFail` log.

Ref https://github.com/zurb/foundation-sites/issues/10828
and https://github.com/gulpjs/gulp/issues/1631

phated:
@sttk what do you think of this change?  It seems like some tools interpret stderr wrong (like `husky`) and treat any output as a failure.  This change is mostly a cosmetic one but it does change some tests.

### DIFF
--- a/test/exports-as-tasks.js
+++ b/test/exports-as-tasks.js
@@ -19,13 +19,13 @@ describe('exports as tasks', function() {
         '--gulpfile ./test/fixtures/gulpfiles/gulpfile-exports.babel.js')
       .run(cb);
 
-    function cb(err, stdout) {
+    function cb(err, stdout, stderr) {
       expect(err).toEqual(null);
-      // Skipping stderr expectation because babel broke node 0.10
-      // expect(stderr).toEqual('');
+      expect(stderr).toEqual('');
       var filepath = path.join(expectedDir, 'tasks-as-exports.txt');
       var expected = fs.readFileSync(filepath, 'utf-8');
-      stdout = eraseTime(skipLines(stdout, 2));
+      // Remove babel/register lines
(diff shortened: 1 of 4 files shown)


sttk:
@phated I think this is good change. I'll review this.

test/exports-as-tasks.js
```diff
@@ -21,11 +21,10 @@ describe('exports as tasks', function() {
 
     function cb(err, stdout) {
       expect(err).toEqual(null);
-      // Skipping stderr expectation because babel broke node 0.10
-      // expect(stderr).toEqual('');
```
sttk: By this pr, stderr should become an empty string. Add its check code, please.

test/flags-require.js
```diff
@@ -50,25 +50,25 @@ describe('flag: --require', function() {
...
+  it('warns if module doesn\'t exist', function(done) {
     runner({ verbose: false })
       .gulp('--require ./null-module.js', '--cwd ./test/fixtures/gulpfiles')
       .run(cb);
 
-    function cb(err, stdout, stderr) {
+    function cb(err, stdout) {
       expect(err).toEqual(null);
-      stderr = eraseLapse(eraseTime(stderr));
-      expect(stderr).toMatch('Failed to load external module ./null-module.js');
```
sttk: By this pr, stderr should become an empty string. Add its check code, please.

phated:
@sttk updated!

sttk:
@phated Perfect! Thanks updating.

sttk:
I've merged.

Issue #146, opened 2018-01-14
Strange behavior of changing working directory.


sttk:
I found a strange behavior about changing working directory. This is related to [the comment](https://github.com/gulpjs/gulp-cli/issues/109#issuecomment-307569185) in #109.

Assuming directory structure:

```
project/
  ├── gulpfile.js
  └── currentdir/
        ├── .gulp.js
        └── gulpfile-2.js
```

and the contents of each files:

```sh
$ cat currentdir/.gulp.js 
module.exports = {
  flags: {
    gulpfile: './gulpfile-2.js'
  }
}

$ cat currentdir/gulpfile-2.js 
var gulp = require('gulp');
gulp.task('default', function() {
  console.log(__filename);
});

$ cat gulpfile.js 
var gulp = require('gulp');
gulp.task('default', function() {
  console.log(__filename);
});
```

In this condition, the following execution uses `project/gulpfile.js` against my intention.

```sh
$ cd currentdir/
$ gulp
[15:36:42] Working directory changed to ~/project
[15:36:42] Using gulpfile ~/project/gulpfile.js
[15:36:42] Starting 'default'...
/Users/sttk/project/gulpfile.js
[15:36:42] Finished 'default' after 141 μs
```

When `project/gulpfile.js` is removed or renamed, `project/currentdir/gulpfile-2.js` is used along my intention.

```sh
$ mv ../gulpfile.js ../gulpfile.js.x
$ gulp 
[15:41:05] Using gulpfile ~/project/currentdir/gulpfile-2.js
[15:41:05] Starting 'default'...
/Users/sttk/project/currentdir/gulpfile-2.js
[15:41:05] Finished 'default' after 95 μs
```

This cause is that gulp found up `project/gulpfile.js` and ~~changed `cwd` to `project/` before finding and applying config files.~~ changed `cwd` for finding a config file to `project/` in liftoff.

~~I think when a gulpfile is specified in a config file, `cwd` should be set back  to `INIT_CWD` or set to parent directory of the gulpfile.~~ I think that `process.INIT_CWD` should be added to [the directories for finding config files](https://github.com/gulpjs/gulp-cli/blob/master/index.js#L45-L52).

----
EDIT: I modified the last two sentences.

phated:
@sttk that is strange, indeed.  Do you think we need to make a change in liftoff?

sttk:
[...]

Issue #95, opened 2016-09-07
local gulp as transitive dependency not found when in a shrinkwrapped direct dependency


JustinTullgren:
Hi,

We have a 'build' npm package that wraps gulp.  This package has two business goals: to control the versions of build dependencies (like gulp, gulp help, gulp debug, less, eslint, webpack, etc) and to provide a common set of tasks to a dependent package.  The tasks are built using convention over configuration but we do allow for extension by exposing the gulp instance and the configuration object we use in the common tasks both as an export and in the return object we send back to callers in the default export.

We have multiple apps that depend on this 'build' package as their sole dependency allowing us to not have to manage multiple packages build dependencies and to allow devs to switch between projects and have a common set of tasks.  

Gulp-cli was able to access the local gulp instance because npm flattens dependencies (including transitive) and gulp was brought in through our 'build' package.  This allowed gulp-cli via Liftoff to find the local gulp directly in node_modules.

However, a few weeks ago a bug was introduced in a patch version of a transitive dependency for webpack (one of our build dependencies) breaking all of our CI and deployment builds as we always do a fresh npm install or npm update.

We decided to finally shrinkwrap our 'build' package before we publish.  However by doing this npm handles the installation of packages differently.  Npm now installs the transitive dependencies in the direct 'build' dependency's node_modules and does not flatten them which means gulp-cli can't find the local gulp.

eg

- node_modules
  - build-package
  - gulp

became

- node_modules
  - build-package
    - node_modules
      - gulp

I understand why this is the desired behavior of npm and that our situation may be atypical and that we may be abusing the use of transitive dependencies vs declaring them as explicit dependencies.

I also understand we could use node directly (`node gulpfile.js`) but then we lose the goodness of what gulp cli give us.

Is there a way for us to provide gulp-cli our local gulp instance inside our gulp file? Exporting it as a const in the gulp file does not work and we don't want to directly have to depend on gulp in each app as we then have to manage our version of gulp in many places and I think we wouldn't be sharing the same instance.

Many thanks!

info:
node 5.8
npm 3.8.2
gulp 3.9.1

phated:
Like you said, this is an npm behavior and we can't really work around it.  You might be able to do some hacky things with the `--cwd` CLI flag but I'm not sure what that would look like.

We are attempting to solve the shared build logic with Custom Registries in gulp 4 (not yet released), but it seems like you are still using 3.9.1 so that doesn't really help the current situation either.

PR #154, opened 2018-01-27
Add initial cwd to directories for searching config files

sttk:
I've add `initCwd` entry to the last of `opts.configFiles`.

This PR is for the issue #146 and changes gulp's behaviors about using config file and selecting gulpfile.

* There is an ancestor directory having `gulpfile.*`.
* And current directory has a config file which has `flags.gulpfile` property.
* And current directory has no gulpfile or a gulpfile of which name is not 'gulpfile.*'.

In the situation above, gulp after this modification uses the config file in current dir and uses gulpfile specified in the config file, but changes cwd to the ancestor dir.
Before this modification, gulp ignores this config file, and finds up a gulpfile in the ancestor dir, changes cwd to the dir and uses the gulpfile.


phated:
@sttk I assume this is going to be changed with the new liftoff stuff?

sttk:
@phated Yes. I think so, too. In addition,  I think we can include a change for `flags.nodeFlags` too with the new liftoff stuff.

phated:
> I think we can include a change for flags.nodeFlags too with the new liftoff stuff.

In separate PRs 😛

sttk:
Sorry. I made a mistake on pushing.

index.js
```diff
@@ -89,7 +93,8 @@ function run() {
     require: opts.require,
     completion: opts.completion,
   }, function(env) {
-    var cfgLoadOrder = ['home', 'cwd'];
+
+    var cfgLoadOrder = ['home', 'cwd', 'initCwd'];
```
phated: @sttk I might be wrong but shouldn't the order be `['home', 'initCwd', 'cwd']`?  I think that `initCwd` is a fallback if we can't find one inside `cwd`, right?

sttk: @phated Perhaps you are right, but [one test case](https://github.com/gulpjs/gulp-cli/pull/154/files#diff-73ba8d51809bb3e4ea85e2ed84a9c49eR80) fails by this because of the bug that `flags.gulpfile` does not change cwd. We should fix the bug before this.

test/config-description.js
```diff
@@ -52,18 +68,55 @@ describe('config: description', function() {
...
+
+    function cb(err, stdout, stderr) {
+      expect(err).toEqual(null);
+      expect(stderr).toEqual('');
+      var expected0 = fs.readFileSync(path.join(expectedDir, 'output0.txt'), 'utf-8');
+      var expected3 = fs.readFileSync(path.join(expectedDir, 'output3.txt'), 'utf-8');
+      stdout = skipLines(eraseTime(stdout), 1);
+
+      expect(headLines(stdout, 1)).toEqual(headLines(expected0, 1));
+      expect(skipLines(stdout, 1)).toEqual(skipLines(expected3, 1));
```
phated: The behavior of this seems extremely awkward.  When using `--cwd`, I wouldn't expect the `flags.gulpfile` to be picking up a gulpfile where I ran my command - I'd expect it to run a gulpfile where my `--cwd` was pointing.

test/config-flags-gulpfile.js
```diff
@@ -51,40 +52,50 @@ describe('config: flags.gulpfile', function() {
...
+      ]);
       done(err);
     }
   });
 
-  it('Should ignore a ./.gulp.* file if another gulpfile is specified by ' +
-  '\n\t--gulpfile', function(done) {
+  // This test case fails because of a bug that `flags.gullpfile` does not
+  // change process.cwd().
+  it.skip('Should use config file here and use gulpfile specified in config file', function(done) {
```
phated: @sttk I just went through this and I think it's too big of issue to ship these changes without having this working correctly.

I think it's a fundamental flaw in how we added config files to Liftoff, because they don't affect it's configBase, etc lookups when we rename the gulpfile.

I'm not sure how to solve this.

phated:
@sttk I'm continuing to dig into this and I think the behavior is even more incorrect when `initCwd` isn't the highest priority.  If I run `cd my-sub-project; gulp` and there's a renamed gulpfile there, my working directory should not be changed to a top-level directory containing a `gulpfile.js` - as most of these tests indicate.

I'm experimenting with `initCwd` being the highest priority and also setting the `env.cwd` if a gulpfile is found in that config.

phated:
@sttk I've also found that when using a `.gulp.*` file to reference a gulpfile in another location, the `.gulp.*` file relative to that gulpfile isn't resolved.  I'm about to push some changes and skipped/failing tests that we need to fix before I think this can land.  I'm also not sure how we can fix it without deeper integrations in Liftoff with the config files.  I'm open to any suggestions.

phated:
@sttk I've just force-pushed a rebase on the lastest master plus my changes where I think our behavior is incorrect.  I've tried to note in comments how things aren't working correctly or the temporary solutions to some problems exist.  Please let me know what you think of all of this.

sttk:
@phated This issue is very difficult. Please give me some time to think.

sttk:
@phated Does what your changes suggest that if flags by config files in `initCwd` has `flags.gulpfile` or `flags.cwd`, `gulp-cli` should change cwd to `env.cwd` and loads `.gulp.*` file in the changed `enc.cwd`?

it means that flags by config files in `initCwd` behaves like flags by CLI options?

sttk:
If so, I have two ideas to fix this issue: (The following codes are not actual but an example.)

1. Process a config file in `initCwd` before `cli.prepare`.

```js
var cli = new Liftoff({ ... });
var foundInInitCwd = fined({ path: '.', name: '.gulp', extensions: cli.extensions });
var cfg = { flags: {} };
if (foundInInitCwd) {
  cfg = require(foundInInitCwd.path);
  opts = mergeConfigToCliFlags(opts, cfg);
  if (!opts.gulpfile) { opts.gulpfile = cfg.flags.gulpfile; }
  if (!opts.cwd) { opts.cwd = cfg.flags.cwd || cfg.flags.gulpfile; }
}

cli.prepare({ cwd: opts.cwd, configPath: opts.gulpfile, ... }, function(env) {
  var cfgLoadOrder = ['home'];
  if (env.configFiles['.gulp']['cwd'].path !== foundInInitCwd.path) { cfgLoadOrder.push('cwd'); }
  var cfg = loadConfigFiles(env.configFiles['.gulp'], cfgLoadOrder);
  opts = mergeConfigToCliFlags(opts, cfg);
  env = mergeConfigToEnvFlags(env, cfg, opts);
   ...
  cli.execute(env, handleArguments);
 });
```

2. Apply [the PR #95](https://github.com/js-cli/js-liftoff/pull/95)

```js
var opts = { ... };
var env = {};
var cli = new Liftoff({
    ...
    configFiles: {
      '.gulp': {
        initCwd: {
          order: 1,
          path: opts.cwd || process.env.INIT_CWD,
          onFound: loadAndMergeConfig, 
        },
        cwd: {  
         order: 2,
          path: env.cwd
          onFound: loadAndMergeConfig, 
        },
        home: { 
          order: 3,
          path: '~',
          onFound: loadAndMergeConfig,
         },
      },
    },
});
cli.prepare({ cwd: opts.cwd, configPath: opts.gulpfile, ... }, function(env) {
   ...
  cli.execute(env, handleArguments);
});

function loadAndMergeConfigs(name, path) { 
  var cfg = require(path);
  opts = mergeConfigToCliFlags(opts, cfg);
  env = mergeConfigToEnvFlags(env, cfg, opts);
}
```

phated:
@sttk I think the issue is actually that too much filesystem traversal is being made in `Liftoff.prototype.buildEnvironment` - I'm wondering if we shouldn't move https://github.com/js-cli/js-liftoff/blob/master/index.js#L62-L111 into the `execute` phase.

The reason I think this is that the `cwd` config file entry in the following code should actually be searching the `process.cwd()` (or `--cwd` or the resolved dirname of `--gulpfile` if either are provided) https://github.com/gulpjs/gulp-cli/blob/1b80d67d6c4c7b9c80a8f49cda009187a91ad88e/index.js#L43-L54

I believe that is what you were trying to achieve by adding `initCwd` config file entry.  If config files were always started at the `cwd` the same way that gulpfiles are looked up, before we attempt to look up the gulpfile, I believe the issue would be solved.  Though, there might be an issue with a `.gulp.*` file existing at a different place in the tree, but traversing up the filesystem should handle that, I think.

sttk:
> I'm wondering if we shouldn't move https://github.com/js-cli/js-liftoff/blob/master/index.js#L62-L111 into the execute phase.

Rather, that moving may be necessary because there is a case that gulp cannot be run only by changing cwd and gulpfile. It's needed to set `env.modulePath` and `env.modulePackage` too by executing those part in the link.

```
$ find . -path './b/node_modules/*' -prune -o -print
.
./a
./a/.gulp.json    => `{ "flags.gulpfile": "../b/gulpfile.js" }`
./a/gulpfile.js    // This is a dummy gulpfile to make gulp-cli load `a/.gulp.json`
./b
./b/gulpfile.js
./b/node_modules
./b/package.json

$ cd a
$ gulp --gulpfile ../b/gulpfile.js
[00:23:11] Working directory changed to ~/path/to/b
[00:23:11] Using gulpfile ~/path/to/b/gulpfile.js
[00:23:11] Starting 'default'...
[00:23:11] Finished 'default' after 2.71 ms

$ gulp
[00:24:57] Local gulp not found in ~/path/to/a
[00:24:57] Try running: npm install gulp
```

sttk:
Furthermore, when the current gulp with `--cwd` option,  it loads and processes `.gulp.*` in the chaged directory.
I think this behavior is correct, but to reproduce the same behavior with a config file it is needed to load and process `.gulp.*` in `initCwd` before resolving paths of other config files.

```
$ find . -path './b/node_modules/*' -prune -o -print
.
./a
./a/.gulp.json    => `{ "flags.gulpfile": "../b/gulpfile.js" }`
./a/gulpfile.js
./b
./b/.gulp.json    => `{ "flags.require": "ansi-colors" }`
./b/gulpfile.js
./b/node_modules
./b/package.json

$ cd a
$ gulp --cwd ../b
[23:38:20] Requiring external module ansi-colors     // => Load and process b/.gulp.json
[23:38:20] Working directory changed to ~/path/to/b
[23:38:20] Using gulpfile ~/path/to/b/gulpfile.js
[23:38:20] Starting 'default'...
[23:38:20] Finished 'default' after 2.4 ms
```

phated:
@sttk I believe this needs conflicts resolved since I merged #239

Issue #157, opened 2018-04-06
`tasks-json`, `version` and `help` flag should imply `silent`


demurgos:
The expected output of `tasks-json` is to emit a JSON representation of the available tasks to the standard output.
Some extra logs may be emitted, causing parse errors by consummers. For example when a gulpfile is transpiled, it emits the line `[<time>] Requiring external module <transpiler>` followed by the JSON output.

The workaround is to use the `--silent` flag to suppress the extra logs. Some tools may be missing this flag so they do not work reliably when extra logs are emitted.

Enabling the `--silent` flag automatically when `--tasks-json` is used and no log-level is configured would fix those errors and improve ergonomics.

- See [related Webstorm issue](https://youtrack.jetbrains.com/issue/WEB-28715)

phated:
I thought we did this.  @sttk any ideas?  @demurgos which version of gulp - the --tasks-json recently was backported to 3.x and maybe we missed the silence part?

sttk:
@demurgos @phated About `--tasks-json`, this issue can be fixed by adding `opts.tasksJson`  in [this condition](https://github.com/gulpjs/gulp-cli/blob/master/lib/shared/log/to-console.js#L35). I'll send a pr for this.

If it is needed to suppress requiring logs for `--version` and `--help` (and more flags?), a few more modifications are needed. (stop logging in first `toConsole`, and enable logging in second `toConsole`).

demurgos:
`--version` and `--help` are intended for humans so the extra logs are less problemematic, but it still would be good to only keep the relevant information. The use of the transpiler is not related to the version or help. I did not look into the code details, but from a user point of view it may be surprising that these commands load the gulpfile.

phated:
:shipit:

phated:
A fix for this has been published in gulp-cli 2.1.0!

PR #159, opened 2018-04-08
Suppress logging when `--tasks-json` flag is specified

sttk:
This pr is to fix the issue #157.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,7 @@
 {
   "extends": "gulp",
   "rules": {
+    "max-len": [1, 90],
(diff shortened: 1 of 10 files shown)


phated:
@sttk I've updated this PR based on your changes that just landed in master.  It's really great to consolidate the `toConsole()` logic!!!

I also added `--help` and `--version` to the suppressed list and added tests for everything.

I had to update your `test/fixtures/config/flags/gulpfile/autoload/.gulp.json` file and related test so I'll have to do that for my changes in #183 if we land this first.

Please review if you have a chance.  Thanks!

Issue #72, opened 2016-04-05
Batch logging to avoid mixing logs from async code


phated:
Ref https://github.com/gulpjs/gulp/issues/361

BurtHarris:
@phated this topic interests me.   Might be something I could help with.   

Is there you got anything written down about logging principles in gulp?   I'm particularly interested in logging from external tools, which might well be running async.   It seems reasonable we might batch up such logging but not display it unless there is an error, and provide hooks for extracting meaningful information from the output.   For example, I've got a tool which says things like 

    Generating file 'C:\code\antlr4ts\src\tree\xpath\.\XPathLexer.ts' for grammar 'XPathLexer.g4'
    Generating file 'C:\code\antlr4ts\src\tree\xpath\.\XPathLexer.tokens' for grammar 'XPathLexer.g4'

Which seems like it might map well into output vinyl files rather than logging output, provided everything works.

BurtHarris:
Seems like this doesn't belong in the **cli** project, am I mistaken?

phated:
The CLI converts messages from https://github.com/gulpjs/gulplog into `console.log`, so yes, it belongs here.

phated:
Relevant code is in gulplog, https://github.com/gulpjs/gulp-cli/blob/master/lib/versioned/%5E4.0.0/log/events.js and https://github.com/gulpjs/gulp-cli/blob/master/lib/shared/log/toConsole.js

radum:
Can we abstract out the logging part out of the CLI one can use whatever logging method one needs?

We could have the default behaviour there, but external logging services could be used to suppress the default one. This will solve this problem and many others I guess.

For me now in particular I'm considering moving away from gulp just because of this where I can't control how the task resolutions are logged.

contra:
@radum Your comment is unrelated to this issue, you should open a new ticket if you have a suggestion. Sounds like you want to be able to overwrite the logger the CLI uses with your own, which is a feature suggestion.

PR #161, opened 2018-04-16
Use theming-log for logging.

sttk:
I've implemented configurable logs with theming-log.

But it is needed [js-liftoff#95](https://github.com/js-cli/js-liftoff/pull/95) to configure logs for require flag and completion flag.


sttk:
I updated this pr based on the current `master` branch.

docs/html/js/write-msg-list.js
```diff
@@ -0,0 +1,117 @@
+'use strict';
```
sttk: Sorry. This file is not used. I'll remove it.

docs/html/js/theme-dic.js
```diff
@@ -0,0 +1,195 @@
...
+    params: [
+      { example: 'This is a bold message.' },
+    ],
+  },
+  code: {
+    desc: 'The function which adds gray background color to a text.',
+    params: [
+      { example: 'This is a code message.' },
+    ],
+  },
```
sttk: Sorry. I mistook about items from `ansi-colors`. There are more items and `code` does not exist.

sttk:
I've modified the above two points about document.

sttk:
This is a sample:

```sh
$ cat .gulp.js
module.exports = {
  log: {
    theme: require('./my-theme'),
    messages: require('./my-msgs'),
  }
};
$ cat my-theme.json 
{
  "ICON": "🥤",
  "VERSION": "{bold: {magenta: {1}}}",
  "TIMESTAMP": "{yellow: [{NOW: YYYY/MM/DD HH:mm:ss}]} ",
  "DESC": "{cyan: {1}}"
}
$ cat my-msgs.json 
{
  "info": {
    "version": "{ICON} {DESC: Gulp version}: {VERSION: {2}}\n{ICON} {DESC: CLI version}: {VERSION: {1}}"
  },
  "error": {
    "gulpfileNotFound": "{ICON} {TIMESTAMP}{ERROR: No gulpfile found}"
  }
}
```
gulp_theming_log

phated:
@sttk I'd like to start work on this again. I noticed the HTML files in this PR - what are those used for?

sttk:
@phated Thanks.
HTML files (and other files in `docs/html`) are sample documents to configure themes and messages. It's not necessary and use as refernece of Gulp document.

sttk:
@phated I created a new branch in my repository, which was based on the current master (v2.2.1), was merged with the change of this PR except `doc/html`, and was modified for #177.

[https://github.com/sttk/gulp-cli/tree/use_theming-log_2](https://github.com/sttk/gulp-cli/tree/use_theming-log_2)

Should I update this PR with that new branch?

phated:
@sttk would you want to include your gulp libraries inside the gulpjs organization?

sttk:
@phated I've updated!

 > would you want to include your gulp libraries inside the gulpjs organization?

Yeah, it's good.
I'd like to include [copy-props](https://github.com/sttk/copy-props), [each-props](https://github.com/sttk/each-props) and [theming-log](https://github.com/sttk/theming-log) into gulp organization.

sttk:
I forgot to test about #177, and did it.

phated:
@sttk thank you! I have confirmed that gulpjs members can create and transfer repositories, so you can transfer those projects when you have time.

phated:
@sttk I believe this needs conflicts resolved since I merged #239

phated:
@sttk are you planning to bring this up-to-date for the gulp-cli v3.0.0 release or should I plan to finish the remaining tasks without color configuration?

sttk:
@phated I'm tackling this issue. Hopefully, I'll finish this weekend.
I hope to include this feature into gulp-cli v3.0.0. Please give me a little more time.

https://github.com/sttk/gulp-cli/commits/ci-test/

phated:
@sttk 👍 sounds good. I will wait for your changes.

sttk:
@phated I've completed the changes for this PR. Please review it.

sttk:
I've added handling for yargs error message.

sttk:
I've made several changes. In Addition, I've removed `config.description` as its purpose overlaps with `config.log.msgs.tasks/tasksJson.description`.

index.js
```diff
@@ -79,26 +81,16 @@ cli.on('loader:success', function(name) {
...
-  }
+  log.warn(msgs.warn.loaderFailure, name, Boolean(error), error.toString());
 });
 
 cli.on('respawn', function(flags, child) {
-  var nodeFlags = chalk.magenta(flags.join(', '));
-  var pid = chalk.magenta(child.pid);
-  log.info('Node flags detected:', nodeFlags);
-  log.info('Respawned to PID:', pid);
+  log.info(msgs.info.respawn, flags.join(', '), child.pid);
```
phated, line 93: These should not be combined, one message is about detecting node flags and one if about the respawn.

index.js
```diff
@@ -79,26 +81,16 @@ cli.on('loader:success', function(name) {
...
 cli.on('loader:failure', function(name, error) {
-  log.warn(
-    chalk.yellow('Failed to load external module:'),
-    chalk.magenta(name)
-  );
-  /* istanbul ignore else */
-  if (error) {
-    log.warn(chalk.yellow(error.toString()));
-  }
+  log.warn(msgs.warn.loaderFailure, name, Boolean(error), error.toString());
```
phated, line 89: I don't like combining the preload message failure and the error printing. Can we keep them separate?

index.js
```diff
@@ -132,69 +124,56 @@ function onExecute(env) {
...
+    makeHelp(parser).showHelp(console.log);
     exit(0);
   }
 
   // Anything that needs to print outside of the logging mechanism should use console.log
   if (env.config.flags.version) {
-    console.log('CLI version:', cliVersion);
-    console.log('Local version:', env.modulePackage.version || 'Unknown');
+    var gulpVersion = env.modulePackage.version || 'Unknown';
+    console.log(format(theme, msgs.info.version, cliVersion, gulpVersion));
```
phated, line 140: I don't think we want to allow this to be configured. We ask for this specific information in our bug reports.

Anywhere we use `console.log` instead of the logger we don't need to support configuration.

lib/shared/completion.js
```diff
@@ -2,21 +2,21 @@
...
 var path = require('path');
+var format = require('theming-log').format;
+
+var theme = require('./log/theme');
+var msgs = require('./log/messages');
 
 module.exports = function(name) {
   if (typeof name !== 'string') {
-    throw new Error('Missing completion type');
+    throw new Error(format(theme, msgs.error.noCompletionType));
```
phated, line 12: Why do we format the message for a thrown error?

lib/shared/completion.js
phated, line 1: This file is for providing shell completions, we probably don't need to use formatting in here at all.

lib/shared/config/merge-configs.js
```diff
@@ -5,6 +5,7 @@ var path = require('path');
 
 function mergeConfigs(env) {
   var cfg = {};
+  /* istanbul ignore if */
```
phated, line 8: Why is this needed now?

lib/shared/log/tasks.js
```diff
@@ -1,7 +1,10 @@
 'use strict';
 
 var log = require('gulplog');
-var chalk = require('chalk');
+var stringWidth = require('string-width');
```
phated, line 4: Why do we need this module?

lib/shared/log/tasks.js
```diff
@@ -12,7 +15,7 @@ function logTasks(tree, opts, getTask) {
   var maxDepth = opts.tasksDepth;
   if (typeof maxDepth !== 'number') {
     maxDepth = 50;
-  } else if (maxDepth < 1) {
+  } else /* istanbul ignore if */ if (maxDepth < 1) {
```
phated, line 18: Why add this comment now?

lib/shared/log/tasks.js
```diff
@@ -29,22 +32,23 @@ function logTasks(tree, opts, getTask) {
...
 function printTaskTree(tree, opts) {
   var lines = [];
-  lines.push({ label: tree.label });
   var maxLabelWidth = 0;
 
   tree.nodes.forEach(function(node, idx, arr) {
     var isLast = idx === arr.length - 1;
     var w = createTreeLines(node, lines, opts, 1, '', isLast);
-    maxLabelWidth = Math.max(maxLabelWidth, w || 0);
+    maxLabelWidth = Math.max(maxLabelWidth, w || /* istanbul ignore next */ 0);
```
phated, line 40: Why add this comment now?

lib/shared/log/tasks.js
```diff
@@ -29,22 +32,23 @@ function logTasks(tree, opts, getTask) {
...
 
+  log.info(msgs.tasks.description, tree.label);
+
   lines.forEach(function(line) {
-    var s = line.label;
     if (line.desc) {
-      var spaces = ' '.repeat(maxLabelWidth - line.label.length) + '  ';
-      s += spaces + line.desc;
+      var spaces = ' '.repeat(maxLabelWidth - line.width) + '  ';
+      log.info(line.fmt, line.bars, line.name, true, spaces, line.desc);
```
phated, line 48: Do you think users will configure this?

lib/shared/log/theme.js
```diff
@@ -0,0 +1,63 @@
...
+  TIMESTAMP: '[{gray: {NOW: HH:mm:ss}}] ',
+
+  IF: function(text) {
+    var idx = text.indexOf('?');
+    var cond = text.substring(0, idx).trim();
+    if (cond === '' || cond === 'false') {
+      return '';
+    }
+    return text.slice(idx + 1);
+  },
```
phated, line 44: Where do we use this? Can we avoid it?

lib/shared/log/timestamp.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
+
+function timestamp(format) {
```
phated, line 3: What is this used for?

lib/shared/log/to-console.js
```diff
@@ -48,10 +58,12 @@ function toConsole(log, opts) {
...
     .forEach(function(level) {
       if (level === 'error') {
-        log.on(level, fancyLog.error);
+        log.on(level, consoleError);
       } else {
-        log.on(level, fancyLog);
+        log.on(level, consoleLog);
       }
+
+      log[level] = themingLog(theme, log[level]);
```
phated, line 66: I'm not sure why you are doing this but I don't think I want to override the logging functions here.

index.js
sttk: It seems that users configuring this would consider these two lines as one.
Additionally, in this way, it is also possible to either output only one line or not output both lines.

index.js
```diff
@@ -132,69 +124,56 @@ function onExecute(env) {
     process.env.UNDERTAKER_SETTLE = 'true';
   }
 
+  if (optsErr) {
+    log.error(msgs.error.failToParseCliOpts, optsErr.message);
+    makeHelp(parser).showHelp(console.error);
+    exit(1);
+  }
```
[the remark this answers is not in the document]
sttk: This part is needed against an error process of yargs for invalid option.
For example, `gulp -f` causes a yargs error because `-f/—gulpfile` requires an argument. By default, yargs outputs the error message `Not enough arguments following: f`, outputs help messages, and exits the program.
To control this behavior and configure these messages, I added this part and [another part](https://github.com/gulpjs/gulp-cli/pull/161/files#diff-1b0b808a2755d08f27702bc502bda7318339b6c8120d9319eec2b6e5e9e00278L125-R81) in `lib/shared/options/parser.js`.

lib/shared/log/tasks.js
sttk: For increasing coverage. However, in the first place,`|| 0` is unnecessary.

lib/shared/config/merge-configs.js
sttk: For increasing coverage. It's impossible to test a config file in user home.

index.js
sttk: Understood. I think that's reasonable.

lib/shared/completion.js
sttk: I also aim to enable multilingualization support with this change. I want to make it possible to configure all messages generated by gulp-cli.

lib/shared/log/tasks.js
sttk: This module is used to calculate the widths of task names or flag names with its configured styles. Those styles might include emojis, symbols, or other full width characters.

lib/shared/log/timestamp.js
sttk: This function is used to display the timestamp at the beginning of logs, similar to fancy-log.
Some users may want to append dates to timestamps, while others may not want to display the timestamp itself for specific messages like #206. There may also be users who want to change the color, as in gulpjs/gulp#2271. To meet such demands, I made displaying timestamp configurable, too.

lib/shared/log/tasks.js
sttk: At least, those colors should be configurable according to console background color.

lib/shared/log/to-console.js
sttk: You're right! This is wrong. This code replaces global gulplog methods, affecting other plugins that use gulplog. Instead of gulplog, we should create our own logger specific to gulp-cli. To use own logger would also eliminate the need to consider the issue #249 (and #254). 
One idea I have is to pass the namespace `Symbol('gulp-cli')` to glogg to create a new logger.

--
This is a digression and a future discussion, but could we potentially resolve #72 by creating loggers by glogg with namespace by task names and plugin names and establishing rules for plugins to use it? It’s still rough idea, and I don’t have thought of specific methods yet.

sttk: I've mistaken. Initialization of the logging functions of gulplog v1 is  necessary, even so.

lib/shared/log/tasks.js
sttk: I'll remove this comment and add a test case to pass it.

sttk:
@phated I've modified what you pointed out except some points still under discussion.

sttk:
@phated I want to add `--no-theme` cli flag and `flags.noTheme` config flag that are not to use log theme and messages in a config file. I believe they could be useful when reporting issues or when not using the theme specified in the project.

Can I add them to this PR?

PR #173, opened 2018-11-27
format-error.js: Prefer `e.error.message` if available

xkr47:
Previously when using `e.error` and it was an object, we got just an "[object Object]" message in the resulting Error object (and error message in terminal). With this patch instead when there is a `message` property inside `e.error`, use that to get a better error report, and falling back to old behaviour in other cases.

### DIFF
--- a/lib/versioned/^3.7.0/format-error.js
+++ b/lib/versioned/^3.7.0/format-error.js
@@ -17,7 +17,7 @@ function formatError(e) {
   }
 
   // Unknown (string, number, etc.)
-  return new Error(String(e.err)).stack;
+  return new Error(String(e.err.message || e.err)).stack;
 }
 
 module.exports = formatError;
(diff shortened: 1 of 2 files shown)


sttk:
@xkr47 Thank you for this PR, but is there a test for verifying this patch? Or are there any examples to reproduce to outputting "[object Object]" message?

phated:
Sounds like your `toString` methods are wrong?

xkr47:
You are right, although it was not exactly "our" toString :)

```js
function Webpack(done) {
  const run        = require('parallel-webpack').run
  const configPath = require.resolve('./webpack/build.js')

  run(configPath, {
    watch               : rawVersion,
    maxRetries          : 0,
    stats               : false,
    maxConcurrentWorkers: 3
  }, done)
}
```

Changed that next-last line to instead do:

```js
  }, (err) => done(err && err.message || err))
```

and now I get errors nicely passed through!

Sorry for the inconvenience. I did not investigate the problem far enough.

Issue #163, opened 2018-05-22
Automatic color detection issue on Windows with GIT Bash


Neonit:
Five minutes ago I didn't even know gulp had color support. I found this by running gulp manually:
`node node_modules/gulp/bin/gulp.js`
This gave me a colored output on my GIT Bash on Windows 7 x64.

The bash itself reports the following version:
```
GNU bash, version 4.3.42(5)-release (x86_64-pc-msys)
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
```

Just running `gulp` (using the globally installed gulp-cli 2.0.1) does **not** give me a colored output. It works when I add `--color` manually. I have not installed gulp globally. My local gulp version is 4.0.0.

Further I'm using Node 10.0.0 and npm 5.6.0.

I would have tried to workaround this by adding the color flag to the global gulp-cli configuration, but the documentation states there is no color flag for the configuration file.

sttk:
Thank you for the report.

I reproduced it but I don't find the cause yet, but I think it is not a gulp issue.
I'll look into it more.

```
$ npm i gulp-cli gulp@4
$ node -v
$ node_modules/.bin/gulp -v   // output with no color.
[05:21:19] CLI version 2.0.1
[05:21:19] Local version 4.0.0

$ node node_modules/gulp/bin/gulp.js -v // output with color.
[05:21:27] CLI version 2.0.1
[05:21:27] Local version 4.0.0

$ node_modules/gulp/bin/gulp.js -v // output with color.
[05:21:36] CLI version 2.0.1
[05:21:36] Local version 4.0.0
```

sttk:
I found the cause of this issue.

gulp-cli uses [color-support](https://www.npmjs.com/package/color-support) to detect whether a terminal supports color output, and it checks whether stdout is a TTY by `process.stdout.isTTY`.

However, this `isTTY` flag on MinGW is `false`, maybe because MinTTY does not play well with Windows console applications which need interactive mode.

We will not solve this issue in gulp-cli.

In additon, `color` flag in configuration file is not supported now.

Please work around this issue by using `alias` command, etc.

Neonit:
I cannot reproduce your findings.

Using the same GIT Bash I encountered the issue with in the directory I cloned gulp-cli to (so I assume it's using the same version of color-support):

```
[...]

PR #176, opened 2018-12-11
Show error info for --require param problem

smialy:
I spent some time to figure out why: `gulp --require @babel/register ...` not works. Was missing `@babel/core` :]

### DIFF
--- a/index.js
+++ b/index.js
@@ -70,7 +70,10 @@ cli.on('require', function(name) {
 
 cli.on('requireFail', function(name, error) {
   var message = name + (error ? ' [' + error + ']' : '');
-  log.warn(ansi.yellow('Failed to load external module'), ansi.magenta(message));
+  log.warn(
+    ansi.yellow('Failed to load external module'),
+    ansi.magenta(message)
+  );
 });
 
 cli.on('respawn', function(flags, child) {
--- a/index.js
+++ b/index.js
@@ -69,11 +69,13 @@ cli.on('require', function(name) {
 });
 
 cli.on('requireFail', function(name, error) {
-  var message = name + (error ? ' [' + error + ']' : '');
   log.warn(
     ansi.yellow('Failed to load external module'),
-    ansi.magenta(message)
+    ansi.magenta(name)
   );
+  if (error) {
+    log.warn(ansi.yellow(error.toString()));
+  }
 });
 
 cli.on('respawn', function(flags, child) {
--- a/index.js
+++ b/index.js
@@ -68,8 +68,9 @@ cli.on('require', function(name) {
   log.info('Requiring external module', ansi.magenta(name));
 });
 
-cli.on('requireFail', function(name) {
-  log.warn(ansi.yellow('Failed to load external module'), ansi.magenta(name));
+cli.on('requireFail', function(name, error) {
+  var message = name + (error ? ' [' + error + ']' : '');
+  log.warn(ansi.yellow('Failed to load external module'), ansi.magenta(message));
 });
 
 cli.on('respawn', function(flags, child) {
--- a/test/exports-as-tasks.js
+++ b/test/exports-as-tasks.js
@@ -25,7 +25,7 @@ describe('exports as tasks', function() {
       var filepath = path.join(expectedDir, 'tasks-as-exports.txt');
       var expected = fs.readFileSync(filepath, 'utf-8');
       // Remove babel/register lines
-      stdout = eraseTime(skipLines(stdout, 3));
+      stdout = eraseTime(skipLines(stdout, 4));
       expect(stdout).toEqual(expected);
       done(err);
     }
--- a/test/fixtures/test-error-module.js
+++ b/test/fixtures/test-error-module.js
@@ -1,0 +1,2 @@
+throw new Error('from error module');
+console.log('inside error module');
--- a/test/flags-require.js
+++ b/test/flags-require.js
@@ -90,4 +90,42 @@ describe('flag: --require', function() {
     }
   });
 
+  it('warns if module throw some error', function(done) {
+    runner({ verbose: false })
+      .gulp('--require ../test-error-module.js', '--cwd ./test/fixtures/gulpfiles')
+      .run(cb);
+
+    function cb(err, stdout, stderr) {
+      stdout = eraseLapse(eraseTime(stdout));
+      expect(err).toEqual(null);
+      expect(stderr).toEqual('');
+      expect(stdout).toMatch('Failed to load external module ../test-error-module.js [Error: from error module]');
+      expect(stdout).toNotMatch('inside error module');
+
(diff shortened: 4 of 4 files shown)


smialy:
I've seen and check in my local machine. All tests pass also on TravisCI too. Only on Appveyor not and it's looks that something is broken there.

phated:
@sttk would this change I made in fancy-log cause this new failure? https://github.com/gulpjs/fancy-log/commit/9a3a6620434ed3eca8fbd6cf0683410ade38c6d4

smialy:
Nice color console require one extra condition: if(testing) { disableColors() } :]

phated:
Interesting.  I'm going to have @sttk chime in as well because I'm guessing I broke something within fancy-log.

smialy:
Some plan or idea how to fix it?

sttk:
@phated The issue you fixed in above link might be similar to #163.
If so, this is not Windows' issue but terminal's (AppVeyor uses MinGW?) or color-support's issue.
But I have no good idea to solve this problem.

Sample code: (color-test.js)

```js
#!/usr/bin/env node
var colorSupport = require('color-support');
console.log('Color support: ', colorSupport());
```

On MinGW:

```dos
$ node color-test.js
Color support:  { level: 1, hasBasic: true, has256: false, has16m: false }

$ node.exe color-test.js
Color support:  false

$ ./color-test.js
Color support:  false
```

On Command Prompt:

```
C:\>node.exe color-test.js
Color support:  { level: 1, hasBasic: true, has256: false, has16m: false }

C:\>node.exe color-test.js
Color support:  { level: 1, hasBasic: true, has256: false, has16m: false }
```

sttk:
@smialy @phated I'll attach `--no-color` flag to gulp command automatically in **gulp-test-tool**.
At least, it will solve this problem about gulp-cli tests.

smialy:
Who can do it?

sttk:
Restart link is not displayed on my account.

@phated Can you restart appveyor test?

phated:
@sttk how can we test that errors are being shown correctly?  I didn't even realize this method had a 2nd argument.

sttk:
@phated By [this line](https://github.com/js-cli/js-liftoff/blob/master/index.js#L35), an error object is passed as 2nd argument.

@smialy Can you write a test for this modification in `test/flag-require.js`?

index.js
```diff
@@ -68,8 +68,9 @@ cli.on('require', function(name) {
   log.info('Requiring external module', ansi.magenta(name));
 });
 
-cli.on('requireFail', function(name) {
-  log.warn(ansi.yellow('Failed to load external module'), ansi.magenta(name));
+cli.on('requireFail', function(name, error) {
+  var message = name + (error ? ' [' + error + ']' : '');
```
sttk: Since `error` is an error object, this code outputs too many lines including stack trace. Isn't it enough error message? And I think it's better that the error message is outputted to next line.

smialy: Fo me this message:
```
Failed to load external module @babel/register (Error: Cannot find module '@babel/core')
```
is more helpful than:
```
Failed to load external module @babel/register
```

sttk: @smialy Sorry, I mistake `console.log(new Error() + '')` for `console.log(new Error())`.
The former does not output stack trace.

sttk: @smialy Now I think this modification is good, but add/modify a test case to `test/flag-require.js`, please.

sttk:
@phated Are you ok with this?

phated:
Should we do some better formatting on this?  I'm thinking something like:
```
Failed to load external module @babel/register
    Encountered error: [Error: Missing @babel/core]
```

Thoughts?

sttk:
@smialy What about the format @phated suggeted?
Could you modify like this?

smialy:
In this way?
```
[09:16:57] Failed to load external module @babel/register
    Error: Cannot find module '@babel/core'
```

sttk:
I think it's enough only the first two lines and it's good except that.

smialy:
Some problem with integration travis-ci and github?

index.js
```diff
@@ -68,8 +68,12 @@ cli.on('require', function(name) {
   log.info('Requiring external module', ansi.magenta(name));
 });
 
-cli.on('requireFail', function(name) {
-  log.warn(ansi.yellow('Failed to load external module'), ansi.magenta(name));
+cli.on('requireFail', function(name, error) {
+  var message = name + (error ? '\n    [' + error + ']' : '');
+  log.warn(
+    ansi.yellow('Failed to load external module'),
+    ansi.magenta(message)
```
sttk: @smialy I'm sorry but I want you to change more.

1. Change color of error message to yellow. Magenta is for module name and this warning message should be yellow because we changed this color to yellow by user's request.
2. Do `log.warn()` each line separately so as to put timestamp at the beginning of each line, like other log outputs.

index.js
```diff
@@ -68,8 +68,14 @@ cli.on('require', function(name) {
   log.info('Requiring external module', ansi.magenta(name));
 });
 
-cli.on('requireFail', function(name) {
-  log.warn(ansi.yellow('Failed to load external module'), ansi.magenta(name));
+cli.on('requireFail', function(name, error) {
+  log.warn(
+    ansi.yellow('Failed to load external module'),
+    ansi.yellow(name)
```
sttk: Sorry, my explanation might be bad. Please modify only a module name  to magenta as is.

phated:
@sttk sorry that I haven't been around to review, I started a new job a couple weeks ago.  Are you satisfied with the PR?  I'll restart the appveyor build but I'm confused by what dependency won't install.

sttk:
The cause of Appveyor's error is bump deps of `normalize-package-data` : `is-builtin-module "^1.0.0"=>"^3.0.0"`. (That was made on Jan. 29).

These modules are used in `yargs`, and  `is-builtin-module` v3.0.0 targets Node >=v6.0.  Since this bumping of `normalize-package-data` is made as a patch (v2.4.0=>v2.4.1), this error cannot be avoided on the older versions of Node.

sttk:
@phated Please restart Appveyor's test because `normalize-package-data`' was changed back to 
 use `is-builtin-module^1.0.0`. The test will succeed.

phated:
Why was this closed?

phated:
@sttk should I publish this version or do you want to land some more stuff first?

phated:
@smialy thanks for all the hard work on this!!!

sttk:
@phated I think it's better to publish `liftoff` with merging the pr [js-liftoff#94](https://github.com/js-cli/js-liftoff/pull/94) before `gulp-cli` because `requireFail` messages will increase by this pr.

phated:
I meant to follow up on this when 2.1.0 was published.  It's been published for 11 days so this fix is available now.  Thanks again!

Issue #181, opened 2019-02-19
.gulp.json not working well with gulpfile.ts


bjg2:
gulp version 4.0.0
gulp CLI version 2.0.1

Please take a look at [gulp-typescript-fail](https://github.com/bjg2/gulp-typescript-fail) repo, there is a small repro.

`gulp test` is not working and `gulp test --gulpfile other_folder/gulpfile.ts` is working, even though the same gulpfile is defined in `.gulp.json`. Also, if I execute `gulp test --gulpfile other_folder/gulpfile.ts` silent flag in `.gulp.json` is completely ignored, while it is working as expected if provided via CLI.

Even if I move `gulpfile.ts` is in the root and I change `.gulp.json` to point at `gulpfile.ts`, silent flag in the `.gulp.json` is working for default messages but is ignored for ts one and `[22:39:59] Requiring external module ts-node/register` is written. The same is not written if it is provided via CLI as `gulp test --silent`.

So the `.gulp.json` is not working for me, am I doing something wrong? How can we fix this?

sttk:
There is a different behavior between `--gulpfile` option and `flags.gulpfile` in `.gulp.js`. `--gulpfile` options changes cwd but `flags.gulpfile` does not. These behaviors are still under consideration.

Maybe you can work that by the following way: 

1. Move `.gulp.json` in `other_folder` and modify `gulpfile.ts` path.
2. Run the command `gulp test --cwd other_folder`. (`flags.cwd` is not supported yet, sorry.)

bjg2:
So it is not possible currently to have configs in `,gulp.json` (not in CLI), write gulpfile in typescript, and to have gulpfiles in other folders. :) Ok, I will align, but I am eagerly waiting for `.gulp.js` flags to be taken as default and CLI ones as override (I guess that is what everyone using `.gulp.json` expects to happen).

sttk:
@bjg2 Sorry, I've mistaken. This problem is not about cwd but about autoloading a gulpfile in `.gulp.js`.

The cause is the lacking of preparing loader for a gulpfile in `.gulp.js`. The same process for a gulpfile by cli option is executed in `liftoff` by using `rechoir`.

I'll try to fix this issue this weekend.

sttk:
> but I am eagerly waiting for `.gulp.js` flags to be taken as default and CLI ones as override (I guess that is what everyone using `.gulp.json` expects to happen).

OK. I'll address that after this issue.

phated:
[...]

PR #182, opened 2019-03-12
Fix the issue #181

sttk:
This PR uses Liftoff v3.1 and moves the loading of config files before the autoloads of modules. 

About adding more config flags and overriding flags from config file by ones from cli args, I will send another PRs.

### DIFF
--- a/index.js
+++ b/index.js
@@ -85,23 +85,29 @@ cli.on('respawn', function(flags, child) {
   log.info('Respawned to PID:', pid);
 });
 
+
 function run() {
-  cli.launch({
+  cli.prepare({
     cwd: opts.cwd,
     configPath: opts.gulpfile,
     require: opts.require,
     completion: opts.completion,
-  }, handleArguments);
+  }, function(env) {
+
+    var cfgLoadOrder = ['home', 'cwd'];
+    var cfg = loadConfigFiles(env.configFiles['.gulp'], cfgLoadOrder);
+    opts = mergeConfigToCliFlags(opts, cfg);
+    env = mergeConfigToEnvFlags(env, cfg);
+    env.configProps = cfg;
+
+    cli.execute(env, handleArguments);
+  });
 }
 
 module.exports = run;
 
 // The actual logic
 function handleArguments(env) {
-  var cfgLoadOrder = ['home', 'cwd'];
-  var cfg = loadConfigFiles(env.configFiles['.gulp'], cfgLoadOrder);
-  opts = mergeConfigToCliFlags(opts, cfg);
-  env = mergeConfigToEnvFlags(env, cfg);
 
   // This translates the --continue flag in gulp
   // To the settle env variable for undertaker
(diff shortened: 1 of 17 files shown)


index.js
```diff
@@ -85,23 +87,29 @@ cli.on('respawn', function(flags, child) {
...
-  }, handleArguments);
+  }, function(env) {
+
+    var cfgLoadOrder = ['home', 'cwd'];
+    var cfg = loadConfigFiles(env.configFiles['.gulp'], cfgLoadOrder);
+    opts = mergeConfigToCliFlags(opts, cfg);
+    env = mergeConfigToEnvFlags(env, cfg);
+    env.configProps = cfg;
+
+    this.execute(env, handleArguments);
```
phated: should this be `cli` instead of `this`? - I'm thinking that will help when we switch to promises.

lib/shared/completion.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+/* eslint no-console: off */
```
phated: Let's just add this to the `.eslintrc` file

lib/versioned/^3.7.0/index.js
```diff
@@ -55,7 +57,7 @@ function execute(opts, env, config) {
     }
     if (opts.tasksJson) {
       tree = taskTree(gulpInst.tasks);
-      if (config.description && isString(config.description)) {
+      if (config.description && typeof config.description === 'string') {
```
phated: Where is `isString` defined?  Was this a bug?

test/fixtures/config/flags/gulpfile/autoload/.gulp.json
```diff
@@ -0,0 +1,6 @@
+{
+  "flags": {
+    "silent": true,
+    "gulpfile": "other_folder/gulpfile-exports.babel.js"
```
phated: this is awesome ❤️

lib/versioned/^3.7.0/index.js
sttk: Yes, this is a bug.

sttk:
@phated I've modified what you pointed out.

> If you want to update your commit message to match our conventional changelog style, I won't need to squash this into master.

I don't want to commit my comment into master. I just wanted to use such word or style of comment. So I'm OK that you squash or modify my commit comment.
But I'll take care of commit comments to match the conventional changelog style for reducing your effort.

phated:
@sttk Thanks for cleaning up the commit messages and my couple of comments.  I force-pushed to your branch with some nitpicks in wording.  I plan to merge this into master as soon as CI finishes and I'm hoping to do some work on outstanding issues today.

phated:
Btw, this will also close #142 because Liftoff was modified to only log if we don't have a success

sttk:
@phated Thank you so much for reviewing and merging!

PR #183, opened 2019-03-13
New: Support `flags.require` config option (closes #108, closes #167)

### LINKED ISSUE #108: Expose `flags.require` as config option
We can make it work very similar to usage with Liftoff by mirroring https://github.com/js-cli/js-liftoff/blob/master/index.js#L118-L125

The `requireLocal` method is exposed on the instance: https://github.com/js-cli/js-liftoff/blob/master/index.js#L28

### LINKED ISSUE #167: [enhancement] more flags to be configurable via a .gulp file
for example the `require `flag

phated:
@sttk please review

### DIFF
--- a/lib/shared/config/env-flags.js
+++ b/lib/shared/config/env-flags.js
@@ -6,6 +6,7 @@ var copyProps = require('copy-props');
 var toFrom = {
   configPath: 'flags.gulpfile',
   configBase: 'flags.gulpfile',
+  require: 'flags.require',
 };
 
 function mergeConfigToEnvFlags(env, config) {
@@ -15,6 +16,9 @@ function mergeConfigToEnvFlags(env, config) {
 function convert(configInfo, envInfo) {
   if (envInfo.keyChain === 'configBase') {
     return path.dirname(configInfo.value);
+  }
+  if (envInfo.keyChain === 'require') {
+    return [].concat(envInfo.value, configInfo.value);
   }
   return configInfo.value;
 }
--- a/test/config-flags-require.js
+++ b/test/config-flags-require.js
@@ -72,20 +72,36 @@ describe('config: flags.require', function() {
     }
   });
 
-  it.skip('missing test for cwd', function(done) {
+  it('resolves relative requires against cwd', function(done) {
     runner
-      .gulp('--cwd flags --gulpfile flags/require/array/gulpfile.js')
+      .gulp('--cwd flags/require/with-cwd')
       .run(cb);
 
     function cb(err, stdout, stderr) {
       expect(err).toEqual(null);
       expect(stderr).toEqual('');
-      expect(eraseTime(stdout)).toEqual(
-        'Requiring external module ./preload_one\n' +
-        'Requiring external module ./preload_two\n' +
-        'preload one!\n' +
-        'preload two!\n' +
-      '');
+      var requiring = eraseTime(headLines(stdout, 1));
+      expect(requiring).toEqual('Requiring external module ../preload');
+      var preload1 = eraseTime(headLines(stdout, 1, 4));
+      expect(preload1).toEqual('hello preload!');
+      done(err);
+    }
+  });
+
+  it('works with absolute paths, ignoring cwd', function(done) {
+    runner
+      .gulp('--cwd flags/require/with-absolute')
+      .run(cb);
+
+    function cb(err, stdout, stderr) {
+      expect(err).toEqual(null);
+      expect(stderr).toEqual('');
+
+      var absolute = path.join(__dirname, './fixtures/config/flags/require/preload');
+      var requiring = eraseTime(headLines(stdout, 1));
+      expect(requiring).toEqual('Requiring external module ' + absolute);
+      var preload1 = eraseTime(headLines(stdout, 1, 4));
+      expect(preload1).toEqual('hello preload!');
       done(err);
     }
   });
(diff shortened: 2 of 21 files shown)


lib/shared/config/env-flags.js
```diff
@@ -16,6 +17,9 @@ function convert(configInfo, envInfo) {
   if (envInfo.keyChain === 'configBase') {
     return path.dirname(configInfo.value);
   }
+  if (envInfo.keyChain === 'require') {
+    return [].concat(envInfo.value, configInfo.value);
+  }
```
sttk: This code should be moved to `load-files.js`, which is executed by each config file.
`env-flags.js` is executed only once for copying configs to `env` after merging config files.

Take notice that if a value of `flags.require` is relative path, it is treated as relative to cwd when requiring. So this value needs to be converted to an absolute path when loading like `flags.gulpfile`.

test/config-flags-require.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+
+var expect = require('expect');
+
+var path = require('path');
+var fixturesDir = path.join(__dirname, 'fixtures/config');
+
+var headLines = require('gulp-test-tools').headLines;
```
sttk: I encountered eslint error "no-unused-vars" here.

lib/shared/config/env-flags.js
phated: Ah, I knew I was probably doing something wrong.  I'll add a test for that too.

phated: @sttk I am looking at this now and I don't think this belongs in `load-files.js` because that would combine the config from `cwd` and `home` config files.  Everything we've done until now has been that `cwd` configuration overrides any configuration in `home`.

I think we should continue that for this configuration, because it would be hard to track where the require was coming from.  Also, it'd be the only option that combines a field from both locations and I'd like to keep everything consistent.

phated: @sttk currently, when using `--cwd` and `--require` on the command-line, you have to be aware that your file will be required relative to `cwd`, is that not the behavior we want to reflect when using the `.gulp.json` file?  Or do we need to fix the `--require` flag as well?

I'm not sure what behavior I would expect from this combination, so I don't know how we should change (if at all).

test/config-flags-require.js
phated: Done.

lib/shared/config/env-flags.js
sttk: > `cwd` configuration overrides any configuration in `home`.

The above code looks not overriding but appending `require` configuration. With no conversion, `load-files.js` overrides configurations. So if overriding, only normalizing `configInfo.value` to array is needed.

sttk: But I think both two cases (overriding and appending) can be requested. For example, a TS user uses `ts-node` ususally so writes requiring it in `~/.gulp.json`, and writes more loaders in `(project-dir)/.gulp.json` by project, but he don't want to use `ts-node` in only one project.

What do you think  about adding one more format to choice overriding or appending like `{ "flags.require": { "path": "module-name", "append": true } }` ?

sttk: About a relative path, I think better that a relative path written in `~/.gulp.*` is relative to home dir because it is looked and can be used by all project, and directory structures of projects are different.

But I notice that what I'm saying can be solved by specifying an absolute path. And certainly, what I'm saying ignores `--cwd/gulpfile` options. Umm...

phated: @sttk Joining `flags.require` from `cwd` and `home` will break people that pull the project from GitHub (for example, your `ts-node` example would break for me if I pulled that down and didn't have the `ts-node` in my `home` `.gulp.js` file) - we'd want to have all the `flags.require` values in the `cwd` `.gulp.js` file.

I think I need to keep this array concat because I want to join the `--require` flags and the `flags.require` property.  This should be the behavior for anything that takes multiple argument values.

> What do you think about adding one more format to choice overriding or appending like { "flags.require": { "path": "module-name", "append": true } } ?

I don't want to max the options more complex like you suggested.

> About a relative path, I think better that a relative path written in ~/.gulp.* is relative to home dir because it is looked and can be used by all project, and directory structures of projects are different. 
> But I notice that what I'm saying can be solved by specifying an absolute path. And certainly, what I'm saying ignores --cwd/gulpfile options. Umm...

I think we should just note that `flags.require` will be resolved to `cwd` unless it's a node module or absolute path.  I can make that change to the docs.

phated:
@sttk I added docs and tests based on our discussions.  Please review again and hopefully we can merge this.

phated:
@sttk you are probably busy, but could you review this so we can land it?

sttk:
@phated Sorry for the delay. I'll review this today. (It's a public holiday in japan today)

phated:
Thanks!  Sorry to interrupt your holiday 😭

lib/shared/config/env-flags.js
sttk: > Joining flags.require from cwd and home will break people that pull the project from GitHub 

Certainly. it might be rare to use `flags.require` in home.

> I think I need to keep this array concat because I want to join the --require flags and the flags.require property. 

I got it. I think this should be so, too.

> I don't want to max the options more complex like you suggested.
I got it.

> I think we should just note that flags.require will be resolved to cwd unless it's a node module or absolute path. I can make that change to the docs.

Thanks for writing docs. I checked it.

phated:
@sttk thanks for reviewing!

PR #184, opened 2019-03-23
Give priority to cli flags higher than flags by config files.

sttk:
This is requested in #181.

### DIFF
--- a/lib/shared/config/cli-flags.js
+++ b/lib/shared/config/cli-flags.js
@@ -13,7 +13,13 @@ var fromTo = {
 };
 
 function mergeConfigToCliFlags(opt, config) {
-  return copyProps(config, opt, fromTo);
+  return copyProps(config, opt, fromTo, defaults);
+}
+
+function defaults(cfgInfo, optInfo) {
+  if (optInfo.value === undefined) {
+    return cfgInfo.value;
+  }
 }
 
 module.exports = mergeConfigToCliFlags;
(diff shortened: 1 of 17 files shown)


test/lib/config-env-flags.js
```diff
@@ -41,16 +41,22 @@ describe('lib: config/env-flags', function() {
...
         silent: false,
         gulpfile: '/path/to/gulpfile',
+        require: ['a', 'b'],
       },
     };
 
-    var result =  mergeConfig(env, config);
+    var opts = {
+      gulpfile: env.configPath,
+      require: 'preloat',
```
phated: Is this a typo?

lib/shared/config/cli-flags.js
```diff
@@ -13,7 +13,13 @@ var fromTo = {
 };
 
 function mergeConfigToCliFlags(opt, config) {
-  return copyProps(config, opt, fromTo);
+  return copyProps(config, opt, fromTo, defaults);
+}
+
+function defaults(cfgInfo, optInfo) {
```
phated: this doesn't seem to do anything

test/lib/config-env-flags.js
phated: I'm going to remove this line because we don't actually use the `opts` like this.  We thread the `require` property through the `env` produced by `liftoff.prepare`

phated:
Thanks @sttk!!

sttk:
@phated Thank you for reviewing, modifying and merging!!

Issue #187, opened 2019-04-06
[feature] Allow Liftoff moduleName to be set with envar.


AndrewLeedham:
I am currently in the process of writing some integration tests for https://github.com/AndrewLeedham/gulp-http2-push-manifest, and I am wanting to run the module against different versions of Gulp (4.0.0 and 3.9.1) to check compatibility.

To achieve this I need 2 local versions of Gulp installed, this can be achieved with yarn aliases (https://yarnpkg.com/en/docs/cli/add#toc-yarn-add-alias), so I could have a package called `gulp-4.0.0` and `gulp-3.9.1`. However gulp-cli only looks for `'gulp'`. It would therefore be useful to specify the moduleName that Liftoff uses to resolve the local Gulp install via an envar (or config option).

I got this working locally by adding:
```javascript
var cli = new Liftoff({
  name: 'gulp',
  moduleName: process.env.GULP_MODULE_NAME || 'gulp'
```
here https://github.com/gulpjs/gulp-cli/blob/master/index.js#L38.

I will create a PR for the envar version, but I am not 100% how the config system works, and whether that would be a preferable approach?

phated:
@AndrewLeedham the plugin interface has always remained the same, why are you testing against both versions of gulp?  Take a look at how we test streams/plugins at https://github.com/gulpjs/vinyl-fs/blob/master/test/dest.js and https://github.com/gulp-sourcemaps/gulp-sourcemaps/blob/master/test/init.test.js

AndrewLeedham:
> @AndrewLeedham the plugin interface has always remained the same, why are you testing against both versions of gulp? Take a look at how we test streams/plugins at https://github.com/gulpjs/vinyl-fs/blob/master/test/dest.js and https://github.com/gulp-sourcemaps/gulp-sourcemaps/blob/master/test/init.test.js

Fair point. There was no specific reason for testing against multiple versions.

PR #188, opened 2019-04-06
New: Allow Liftoff moduleName to be set via envar

AndrewLeedham:
Fixes: #187


phated:
I don't think this is the correct solution but I'm afk and will need to do more research.

AndrewLeedham:
@phated Fair enough. Do you mean changing the `moduleName` or just using an envar instead of a config option or flag?

PR #189, opened 2019-04-07
Fix: Honor displayName in ESM exports.

coreyfarrell:
This reimplements #53 and expands the regression test to include
parallel tasks.

Fixes gulpjs/gulp#2270


phated:
This looks great!  I'm going to label it as a "new" feature though 😛

PR #190, opened 2019-04-18
Add tests and raise coverage

sttk:
* Change tests of log-level by config files which executed verifying dependencies and always failed offline.
* Add some tests and istanbul ignore comments, and raise coverage to 100%.


index.js
```diff
@@ -186,9 +192,10 @@ function handleArguments(env) {
   var range = findRange(env.modulePackage.version, ranges);
 
   if (!range) {
-    return log.error(
+    log.error(
       ansi.red('Unsupported gulp version', env.modulePackage.version)
     );
+    exit(1);
```
phated: I like this a lot more than return 😄

lib/shared/config/env-flags.js
```diff
@@ -34,11 +34,10 @@ function mergeConfigToEnvFlags(env, config, cliOpts) {
       return [].concat(envInfo.value, configInfo.value);
     }
 
+    /* istanbul ignore else */
     if (envInfo.keyChain === 'nodeFlags') {
-      return [].concat(configInfo.value || []);
+      return [].concat(configInfo.value);
```
phated: If `configInfo.value` is `undefined`, an `undefined` value will be concat'ed into the array.

phated: We should add a test for that.

test/flags-version.js
```diff
@@ -42,4 +43,20 @@ describe('flag: --version', function() {
...
+    runner({ verbose: false })
+      .gulp('--version', '--cwd', os.tmpdir())
+      .run(cb);
+
+    function cb(err, stdout, stderr) {
+      expect(err).toEqual(null);
+      expect(stderr).toEqual('');
+      expect(stdout).toEqual(
+        'CLI version ' + cliVersion + '\n' +
+        ''
```
phated: Should this instead print `Local version: Unknown` instead of just being blank?

sttk: I modified to print `Local version: Unknown`. And I remove the check if `env.modulePackage` is not nullish, because js-liftoff convert it from `null` to `{}` at [this line](https://github.com/js-cli/js-liftoff/blob/master/index.js#L138).

sttk:
@phated I've modified and added tests.

index.js
```diff
@@ -122,20 +123,20 @@ function handleArguments(env) {
 
   // Anything that needs to print outside of the logging mechanism should use console.log
   if (opts.version) {
-    console.log('CLI version', cliVersion);
-    if (env.modulePackage && typeof env.modulePackage.version !== 'undefined') {
-      console.log('Local version', env.modulePackage.version);
-    }
+    console.log('CLI version:', cliVersion);
+    console.log('Local version:', env.modulePackage.version || 'Unknown');
```
phated: I'm pretty sure there is the chance that `env.modulePackage` can be undefined, so we need to keep a check for that. It'll likely need an `/* istanbul ignore */` comment

test/flags-version.js
phated: @sttk haha, I didn't know that!

lib/shared/config/env-flags.js
phated: Looks like a test was added.

phated:
Awesome work @sttk!!  Thanks for taking care of this 😄

sttk:
@phated Thank you for reviewing and merging!

PR #197, opened 2019-09-25
bump dependencies, remove unmaintained node versions

devoidfury:
The main goal of this PR is to get rid of scary-looking warnings from npm audit and do a little house-cleaning on the deps. There's still a few more in `gulp-test-tools`, but I sent a PR over there as well and if that works out, we can bump the package.json over here to reflect that and clear up all those warnings.

Some of the newer versions of dependencies (mocha, for example) dropped support for older unmaintained node versions (<6), so that's been pulled from the CI builds and indicated in the package.json as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,10 @@
 sudo: false
 language: node_js
 node_js:
+  - "12"
   - "10"
   - "8"
   - "6"
-  - "4"
-  - "0.12"
-  - "0.10"
 after_script:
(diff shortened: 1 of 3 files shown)


phated:
npm audit is a terrible tool, you should disable it.

Gulp 4.x supports back to node 0.10, and any changes require a major bump as you would be breaking people. When gulp 5 development begins, these versions (and others) will be dropped.

PR #201, opened 2019-10-29
Use path.join to concatenate paths to support Windows extended-length paths

### LINKED ISSUE #200: Gulp CLI fails with Windows Extended-Length Path
If Gulp is invoked with a [Windows Extended-Length Path](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#maximum-path-length-limitation), then it immediately fails with an `ENOENT` error:

```powershell
> node \\?\C:\Users\chuck\github\gulp-cli\bin\gulp.js -v
internal/fs/utils.js:220
    throw err;
    ^

Error: ENOENT: no such file or directory, scandir '\\?\C:\Users\chuck\github\gulp-cli/lib/versioned/'
    at Object.readdirSync (fs.js:854:3)
    at Object.<anonymous> (\\?\C:\Users\chuck\github\gulp-cli\index.js:31:17)
    at Module._compile (internal/modules/cjs/loader.js:956:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:973:10)
    at Module.load (internal/modules/cjs/loader.js:812:32)
    at Function.Module._load (internal/modules/cjs/loader.js:724:14)
    at Module.require (internal/modules/cjs/loader.js:849:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (\\?\C:\Users\chuck\github\gulp-cli\bin\gulp.js:5:1)
    at Module._compile (internal/modules/cjs/loader.js:956:30) {
  errno: -4058,
  syscall: 'scandir',
  code: 'ENOENT',
  path: '\\\\?\\C:\\Users\\chuck\\github\\gulp-cli/lib/versioned/'
}
```
This appears to be because a path is created using simple string concatenation, instead of `path.join()` in [this file](https://github.com/gulpjs/gulp-cli/blob/d0cf3126fc76ba96e08b09d9df66b4d8b7c99550/index.js#L31). Unlike normal paths, extended-length paths in Windows don't support `/` as a separator, so the resulting string is an invalid path when `__dirname` is an extended-length path.

charlespierce:
Closes #200 

Info
-----
Windows extended-length paths don't support `/` as a path separator, so building a path using string concatenation will fail if the path is extended-length. This causes gulp to fail if it is invoked with an extended-length path, since that causes `__dirname` to be extended-length.

Changes
-----
Updated the path building in `index.js` to use `path.join()`

Tested
-----
Invoked `gulp -v` using an extended-length path and verified that it runs correctly and doesn't throw an error.

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,7 +28,7 @@ var logVerify = require('./lib/shared/log/verify');
 var logBlacklistError = require('./lib/shared/log/blacklist-error');
 
 // Get supported ranges
-var ranges = fs.readdirSync(__dirname + '/lib/versioned/');
+var ranges = fs.readdirSync(path.join(__dirname, '/lib/versioned/'));
 
 // Set env var for ORIGINAL cwd
 // before anything touches it


phated:
This is my favorite kind of bug report (especially coming with a fix!).

Do you know of any way that we can test for this regression on appveyor?

charlespierce:
@phated I'm not sure of a way to run the command in AppVeyor. I tested it locally by running `node \\?\C:\path\to\gulp-cli\bin\gulp-cli.js`, so it would likely need something similar. I don't know if it's possible to get an extended-length path from AppVeyor settings, however.

phated:
I'm going to get this merged for the next release. I'm really sorry it stalled for so long (lots os stuff has been happening in life 😬).

Ideally, I'd like to get a regression test added but I'll open an issue for that and CC you. Thanks for your contribution!

phated:
Thanks for hanging in there! I know it's hard to stay motivated when we take awhile to respond. I'm going to get this out tonight once CI finishes.

PR #204, opened 2020-04-16
Fix false positive reporting of sync tasks on errors

### LINKED ISSUE #162: Clear "task did not complete" warning when an error occurs
Ref https://github.com/gulpjs/gulp/issues/2168

### LINKED ISSUE #203: Message about non-completed tasks is confusing on failure
I'm using `gulp.parallel` to run several tasks in parallel. When the build succeed, everything is fine and gulp is happy with tasks completing. But when one of the tasks is failing (due to an error in one of my sass files for instance), I get a confusing output, after the plugin error, saying that some tasks failed to complete (basically all other parallel tasks, unless they were already done) and the question `Did you forget to signal async completion?`.

I know that this message is meant to report issues with sync tasks. But it seems to me that the implementation should take into account the case of failures, where it is entirely normal than some tasks don't complete, when `gulp.parallel` was used (maybe `gulp.series` suffers from it as well).

stof:
When a failure happens in a parallel task, it is totally normal that other tasks have not reported completion, as gulp.parallel will report the error without waiting for other tasks.

Closes #203 
Closes #162

### DIFF
--- a/test/fixtures/gulpfiles/gulpfile-parallel-failure.js
+++ b/test/fixtures/gulpfiles/gulpfile-parallel-failure.js
@@ -1,0 +1,18 @@
+'use strict';
+
+var gulp = require('gulp');
+
+function noop(cb) {
+  cb();
+}
+
+function errorFunction(cb) {
+  cb(new Error('Error!'));
+}
+
+function notCompleting1() {
+  // Callback is not called
+}
+
+gulp.task('default', gulp.parallel(errorFunction, noop));
+gulp.task('broken', gulp.parallel(errorFunction, noop, notCompleting1));
--- a/lib/versioned/^4.0.0/log/sync-task.js
+++ b/lib/versioned/^4.0.0/log/sync-task.js
@@ -35,12 +35,18 @@ function clear(e) {
   delete tasks[e.uid];
 }
 
+function clearAll() {
+  tasks = {};
+}
+
 function logSyncTask(gulpInst) {
 
   process.once('exit', warn);
   gulpInst.on('start', start);
   gulpInst.on('stop', clear);
-  gulpInst.on('error', clear);
+  // When not running in --continue mode, we need to clear everything on error to avoid
+  // false positives.
+  gulpInst.on('error', process.env.UNDERTAKER_SETTLE === 'true' ? clear : clearAll);
 }
 
 module.exports = logSyncTask;
--- a/test/non-completing-tasks.js
+++ b/test/non-completing-tasks.js
@@ -33,4 +33,31 @@ describe('sync-task', function() {
         done();
       });
   });
+
+  it('should not log false positive in case of parallel failure', function(done) {
+    runner({ verbose: false })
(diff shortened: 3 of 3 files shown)


test/fixtures/gulpfiles/gulpfile-parallel-failure.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+var gulp = require('gulp');
+
+function noop(cb) {
+  cb();
+}
+
+function errorFunction() {
+  throw new Error('Error!');
```
terinjokes, line 10: you're not completing this task, `cb(new Error('Error!'))`.

phated: This should be fine, the domains will catch it.

phated:
This solution seems solid but give me a bit to think on it.

Should be able to ship this weekend.

test/fixtures/gulpfiles/gulpfile-parallel-failure.js
terinjokes: That's what I thought too. Our documentation makes it clear synchronous tasks aren't supported, and throwing an error is technically synchronous.

stof: I changed the test. But the effect is the same. An exception thrown synchronously is already catched and considered as a failure completion.

stof:
test test failure on node 0.10 looks unrelated to my change.

stof:
and indeed, the latest build on master has even more such timeout failures.

stof:
@phated did you have a chance to think about it ?

phated:
@stof thanks for the reminder! Actually, something came to mind this morning: I think we need to support the `--continue` flag within this change. If someone specifies the `--continue` flag, we want to continue to report errors and sync tasks because the CLI will won't exit on the first error. Does that make sense?

I'm not sure how hard it will be to thread that flag through the system. It can be specified in the config file too, so we have to thread it from the main function.

stof:
I would prefer if you could work on adding support for `--continue` yourselves. I fear I don't know the project good enough to figure out how to get this flag value.

stof:
@phated AFAIU, even in `--continue` mode, there is no guarantee that `stop` or `error` will be called for each task. Errors are delayed until everything is settled. But then, as single reporting is done in case of an error.

stof:
@phated PR updated to account for `--continue`, with 2 new tests to ensure we don't regress on these.
Fortunately, detecting this was quite simple, as the sync-tasks logic runs after that `process.env.UNDERTAKER_SETTLE` is updated based on all possible config, so I could read it (as undertaker does), without having to pass it around explicitly.

phated:
@stof thanks, looking at this today.

stof:
@phated any news on this ?

phated:
@stof I've been really swamped recently. Sorry about that.

I will try to find some time coming up. Thanks for the ping.

phated:
Thanks for hanging in there! I know it's hard to stay motivated when we take awhile to respond. I'm going to get this out tonight once CI finishes.

PR #214, opened 2020-05-27
Add ESM support

snoack:
See gulpjs/interpret#65.

### DIFF
--- a/lib/shared/require-or-import.js
+++ b/lib/shared/require-or-import.js
@@ -1,0 +1,28 @@
+'use strict';
+
+var pathToFileURL = require('url').pathToFileURL;
+
+var importESM;
+try {
+  importESM = new Function('id', 'return import(id);');
+} catch (e) {
+  importESM = null;
+}
+
+function requireOrImport(path, callback) {
+  var err = null;
+  var cjs;
+  try {
+    cjs = require(path);
+  } catch (e) {
+    if (pathToFileURL && importESM && e.code === 'ERR_REQUIRE_ESM') {
+      var url = pathToFileURL(path);
+      importESM(url).then(function(esm) { callback(null, esm); }, callback);
+      return;
+    }
+    err = e;
+  }
+  process.nextTick(function() { callback(err, cjs); });
+}
+
+module.exports = requireOrImport;
--- a/lib/versioned/^4.0.0/index.js
+++ b/lib/versioned/^4.0.0/index.js
@@ -17,6 +17,7 @@ var registerExports = require('../../shared/register-exports');
 
 var copyTree = require('../../shared/log/copy-tree');
 var getTask = require('./log/get-task');
+var requireOrImport = require('../../shared/require-or-import');
 
 function execute(opts, env, config) {
 
@@ -33,16 +34,18 @@ function execute(opts, env, config) {
   logSyncTask(gulpInst, opts);
 
   // This is what actually loads up the gulpfile
-  var exported = require(env.configPath);
+  requireOrImport(env.configPath, function(err, exported) {
+    if (err) {
+      console.error(err);
+      exit(1);
+    }
 
-  registerExports(gulpInst, exported);
+    registerExports(gulpInst, exported);
 
-  // Always unmute stdout after gulpfile is required
-  stdout.unmute();
+    // Always unmute stdout after gulpfile is required
+    stdout.unmute();
 
-  process.nextTick(function() {
     var tree;
-
     if (opts.tasksSimple) {
       tree = gulpInst.tree();
       return logTasksSimple(tree.nodes);
--- a/lib/versioned/^3.7.0/index.js
+++ b/lib/versioned/^3.7.0/index.js
@@ -11,9 +11,11 @@ var copyTree = require('../../shared/log/copy-tree');
 var tildify = require('../../shared/tildify');
 var logTasks = require('../../shared/log/tasks');
 var ansi = require('../../shared/ansi');
+var exit = require('../../shared/exit');
 var logEvents = require('./log/events');
 var logTasksSimple = require('./log/tasks-simple');
 var registerExports = require('../../shared/register-exports');
+var requireOrImport = require('../../shared/require-or-import');
 
 function execute(opts, env, config) {
   var tasks = opts._;
@@ -25,20 +27,23 @@ function execute(opts, env, config) {
   }
 
   // This is what actually loads up the gulpfile
-  var exported = require(env.configPath);
-  log.info('Using gulpfile', ansi.magenta(tildify(env.configPath)));
+  requireOrImport(env.configPath, function(err, exported) {
+    if (err) {
+      console.error(err);
+      exit(1);
+    }
 
-  var gulpInst = require(env.modulePath);
-  logEvents(gulpInst);
+    log.info('Using gulpfile', ansi.magenta(tildify(env.configPath)));
 
-  registerExports(gulpInst, exported);
+    var gulpInst = require(env.modulePath);
+    logEvents(gulpInst);
 
-  // Always unmute stdout after gulpfile is required
-  stdout.unmute();
+    registerExports(gulpInst, exported);
 
-  process.nextTick(function() {
+    // Always unmute stdout after gulpfile is required
+    stdout.unmute();
+
     var tree;
-
     if (opts.tasksSimple) {
       return logTasksSimple(env, gulpInst);
     }
--- a/lib/versioned/^4.0.0-alpha.1/index.js
+++ b/lib/versioned/^4.0.0-alpha.1/index.js
@@ -16,6 +16,7 @@ var logTasksSimple = require('../^4.0.0/log/tasks-simple');
 var registerExports = require('../../shared/register-exports');
 
 var copyTree = require('../../shared/log/copy-tree');
+var requireOrImport = require('../../shared/require-or-import');
 
 function execute(opts, env, config) {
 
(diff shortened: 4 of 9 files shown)


.eslintignore
```diff
@@ -1,2 +1,3 @@
 **/*.babel.js
 coverage
+lib/shared/require-or-import-es6.js
```
snoack, line 3: Dynamic `import()` is not yet recognized by ESLint's built-in parser. However, if you want we can lint this file using `babel-eslint`?

lib/versioned/^4.0.0/index.js
```diff
@@ -33,16 +34,18 @@ function execute(opts, env, config) {
   logSyncTask(gulpInst, opts);
 
   // This is what actually loads up the gulpfile
-  var exported = require(env.configPath);
+  requireOrImport(env.configPath, function(err, exported) {
+    if (err) {
+      console.error(err);
```
snoack, line 39: Previously, if `require()` failed we got an unhandled exception on the module level. So I use `console.error()`+`exit()` here to mimick the old behavior as close as possible, or would you rather log this error differently? (I noticed that you generally seem to use `gulplog`).

Also I'm aware that I still have to do similar changes to the other `lib/versioned/*/index.js` scripts. But I'd first like to get your feedback on the general approach before moving forward. What do you think?

lib/shared/require-or-import.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+try {
+  module.exports = require('./require-or-import-es6');
```
snoack, line 4: Node.js <10 errors out with a SyntaxError when loading a script that uses `import()`. By adding another layer and falling back to a `require()`-only implementation (below), we can keep supporting all versions all the way back to Node.js 0.10.

.eslintignore
snoack: Never mind. I simplified the approach, and as a side effect this is no longer an issue.

lib/shared/require-or-import.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+var importESM;
+try {
+  importESM = new Function('id', 'return import(id);');
+} catch (e) {
+  importESM = null;
+}
```
snoack, line 8: Node.js <10 errors out with a SyntaxError when loading a script that uses `import()`. So I dynamically create a function for that and catch the SyntaxError. That way we can keep supporting all Node.js versions all the way back to 0.10.

lib/versioned/^4.0.0/index.js
```diff
@@ -33,16 +34,18 @@ function execute(opts, env, config) {
...
 
-  registerExports(gulpInst, exported);
+    registerExports(gulpInst, exported);
 
-  // Always unmute stdout after gulpfile is required
-  stdout.unmute();
+    // Always unmute stdout after gulpfile is required
+    stdout.unmute();
 
-  process.nextTick(function() {
```
snoack, line 43: Do you know why this was added? I'm merely guessing that it's no longer required as this code is now in the asynchronous callback to `requireOrImport()`. Would you agree?

index.js
```diff
@@ -39,7 +39,14 @@ var cli = new Liftoff({
   processTitle: makeTitle('gulp', process.argv.slice(2)),
   completions: completion,
   extensions: interpret.jsVariants,
-  v8flags: v8flags,
+  v8flags: function(callback) {
+    v8flags(function(err, flags) {
+      if (!err) {
+        flags.push('--experimental_modules');
```
snoack, line 45: This isn't a v8 flag but a Node.js specific flag, so it isn't discovered by `v8flags`. This flag is required in order to use ESM modules before Node.js 13.2.0.

Silic0nS0ldier: And before 12.17.0 (https://nodejs.org/en/blog/release/v12.17.0/), for those not aware.

snoack: I submitted a pull request, making `v8flags` detect non-v8 flags like this: gulpjs/v8flags#55

lib/shared/require-or-import.js
```diff
@@ -0,0 +1,25 @@
...
+
+function requireOrImport(id, callback) {
+  var err = null;
+  var cjs = null;
+  try {
+    cjs = require(id);
```
ExE-Boss, line 14:
```suggestion
    var cjs = require(id);
    process.nextTick(function () {
      callback(null, cjs);
    });
    return;
```

lib/shared/require-or-import.js
```diff
@@ -0,0 +1,25 @@
...
+  } catch (e) {
+    if (importESM && e.code === 'ERR_REQUIRE_ESM') {
+      importESM(id).then(function(esm) { callback(null, esm); }, callback);
+      return;
+    }
+    err = e;
```
ExE-Boss, line 20:
```suggestion
    process.nextTick(function () {
      callback(e);
    });
```

lib/shared/require-or-import.js
```diff
@@ -0,0 +1,25 @@
...
+var importESM;
+try {
+  importESM = new Function('id', 'return import(id);');
+} catch (e) {
+  importESM = null;
+}
+
+function requireOrImport(id, callback) {
+  var err = null;
+  var cjs = null;
```
ExE-Boss, line 12:
```suggestion
```

lib/shared/require-or-import.js
```diff
@@ -0,0 +1,25 @@
...
+  try {
+    cjs = require(id);
+  } catch (e) {
+    if (importESM && e.code === 'ERR_REQUIRE_ESM') {
+      importESM(id).then(function(esm) { callback(null, esm); }, callback);
+      return;
+    }
+    err = e;
+  }
+  process.nextTick(callback.bind(null, err, cjs));
```
ExE-Boss, line 22:
```suggestion
```

ExE-Boss, line 22: This removes the observable difference in `this`, `arguments.length` and `arguments[1]` when `import(…)` is used instead of `require(…)` and an error occurs:

```suggestion
  var cjs;
  try {
    cjs = require(id);
  } catch (err) {
    if (importESM && err.code === 'ERR_REQUIRE_ESM') {
      importESM(id).then(function(esm) { callback(null, esm); }, callback);
      return;
    }
    process.nextTick(function() {
      callback(err);
    });
    return;
  }
  process.nextTick(function () {
    callback(null, cjs);
  });
```

---

### Current:

| callback parameter | `require(…)`      | `import(…)`
| ------------------ | ------------      | -----------
| `this`             | `null`            | `undefined`
| `error`            | `Error \| null`   | `Error \| null`
| `imports`          | `exports \| null` | `exports \| undefined`
| `arguments.length` | `2`               | `1 \| 2`

### With this:

| callback parameter | `require(…)`           | `import(…)`
| ------------------ | ------------           | -----------
| `this`             | `undefined`            | `undefined`
| `error`            | `Error \| null`        | `Error \| null`
| `imports`          | `exports \| undefined` | `exports \| undefined`
| `arguments.length` | `1 \| 2`               | `1 \| 2`

snoack: Thanks for pointing this out, but I wonder if it's worth the added complexity and duplication required to address this, in particular since this is an internal API. (FWIW, if it wouldn't be for Node.js 0.10 support, I would just create and return a new `Promise` here.) @phated, what do you think?

phated: I think the duplication is probably alright for this release. We are definitely switching everything to promises for the next major, but I really like that we can "backport" this to support our current release stream. That's why I'm so excited you put this together!

snoack: Sure, I'm not suggesting to drop Node.js 0.10 support for this release. But I'm wondering whether you agree with @Exe-Boss or if you think my current implementation is fine? I mean it works, and there is no way a user can even tell the difference since it is an internal API. So I'm kinda reluctant to add redundant logic just for some hypothetically case. But your call. :)

phated: Sorry, I haven't fully reviewed this yet, so I was just responding to the questions.

I'll keep this in mind when I review.

index.js
snoack: I removed this workaround and required `v8flags ^3.2.0` which includes my change instead.

lib/shared/require-or-import.js
snoack: I partially addressed this now, except for the `arguments.length` case as this would require additional boilerplate (and I rather not bloat the code for a purely theoretical case).

lib/shared/require-or-import.js
```diff
@@ -0,0 +1,28 @@
...
+}
+
+function requireOrImport(path, callback) {
+  var err = null;
+  var cjs;
+  try {
+    cjs = require(path);
+  } catch (e) {
+    if (pathToFileURL && importESM && e.code === 'ERR_REQUIRE_ESM') {
+      var url = pathToFileURL(path);
```
snoack, line 19: I just added this as I realized that `import()` fails if providing a file path on Windows.

lib/versioned/^4.0.0/index.js
phated: I think this works well! Good solution :)

phated:
This is absolutely amazing! Thank you for all the work you've done here 🙇 

I'm going to get this merged and add some of your comments from GitHub into the source for "future us"

I believe I also need to add the `.mjs` extension to the 1.x and 2.x release streams of interpret.

phated:
Sorry this is taking so long. I didn't realize getting the `.mjs` extension supported (and tested) properly in `interpret` would take so long. I'm hoping to have this released before I close my laptop for the night.

phated:
I just pushed up the remaining changes that needed to land for this to work with `interpret`/`rechoir`/`liftoff` and autodetect the extension. However, CI needs to run and it's super late for me, so I will have to get this released tomorrow afternoon.

Sorry for the delay.

phated:
Running into some issues with ESM support on node 10 for Windows. Need to dig into that.

snoack:
Do you have more details? Tests pass on my Windows machine.

phated:
@snoack https://ci.appveyor.com/project/gulpjs/gulp-cli/builds/33284247

phated:
Of course I create a branch and start debugging and it starts working 🤷.

snoack:
FWIW, I had no luck reproducing it locally, even with Node.js 10.20.1 (32-bit), the exact same version you got on Appveyor.

For reference, a quick web search for 3221225477 (the exit status Node.js was failing with on Appveyor) indicates some kind of crash in Windows likely due to environmental issues.

phated:
Thanks for digging in! It definitely seems like a weird issue with AppVeyor. Anyway, this has just been released as v2.3.0! Thanks for all your work here 🎉

PR #216, opened 2020-07-13
Add install section to readme

SeekingMeaning:
Homebrew page: https://formulae.brew.sh/formula/gulp-cli

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,20 @@
 
 Command Line Utility for Gulp
 
+## Install
+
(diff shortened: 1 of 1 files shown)


phated:
We don't control homebrew so I don't want to recommend it. We also have installation instructions on our homepage docs so I'm not sure this is necessary.

I'll merge though if you remove the homebrew instructions.

phated:
Thanks!

PR #217, opened 2020-08-19
Fix ERR_UNSUPPORTED_DIR_IMPORT for gulpfile.mjs

ngdangtu-vn:
### What is this PR for?
This is a ERR_UNSUPPORTED_DIR_IMPORT fixed patch. (And did a bit code refactor in `require-or-import.js` file)

Before this patch, gulp-cli won't be able to read directory URL for ESM mode. For example:
```TXT
/gulpfile.js		# work
/gulpfile.js/index.js	# work
/gulpfile.mjs		# work
/gulpfile.mjs/index.mjs	# fail
```
The only solution is append `index.mjs` to the directory. And I fixed it.

### Was it tested yet?
Almost everything.

- [x] It should not break anything -> Tested with mocha testcases (available in this repo)
- [x] It should be able to read `gulpfile.mjs` directory -> Tested directly in my own repo (a)
- [ ] A mocha testcase for reading `gulpfile.mjs` directory -> I have no idea how to do it, so this task is under researching

### DIFF
--- a/lib/shared/require-or-import.js
+++ b/lib/shared/require-or-import.js
@@ -1,6 +1,8 @@
 'use strict';
 
 var pathToFileURL = require('url').pathToFileURL;
+var pathStatsSync = require('fs').lstatSync;
+var join = require('path').join;
 
 var importESM;
 try {
@@ -12,6 +14,28 @@ try {
   importESM = null;
 }
 
+function securePath(path) {
+  var pathStatsSync = require('fs').lstatSync;
+  var join = require('path').join;
+
+  // Avoid ERR_UNSUPPORTED_DIR_IMPORT (>=NodeJs 12)
+  var validPath = pathStatsSync(path).isDirectory() ?
+    join(path, 'index.mjs') :
+    path;
+
+  // This is needed on Windows, because import() fails if providing a Windows file path.
+  return pathToFileURL(validPath);
+}
+
+function onImportESM(path, callback) {
+  var safePath = securePath(path);
+  importESM(safePath)
+    .then(function(modules) {
+      callback(null, modules);
+    })
+    .catch(callback);
+}
+
 function requireOrImport(path, callback) {
   var err = null;
   var cjs;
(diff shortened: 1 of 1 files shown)


ngdangtu-vn:
#### Import directory testcase mentioned above:

(a) My testcase was simple, I create an `index.mjs` inside `gulpfile.mjs`:
```TXT
// repo structure
/
	.d/
	src/
		gulpfile.mjs/
			index.mjs
	a/
		files.txt
	b/
		x.conf
	c.txt
```

```Js
// gulpfile.mjs/index.mjs file
import gulp from 'gulp'
import path from 'path'

const dir = [
	path.join('..', 'a'),
	path.join('..', 'b/*.conf'),
	path.join('..', 'c.txt')
]

const dest = '../.d/'

export function copy() {
	return gulp.src(theme, { base: '..' })
		.pipe(gulp.dest(dest))
}

export default gulp.series(cpTheme)
```

From `/`, I `cd src` and then `gulp copy`. The result work as it should be, all the files/dir move to the `/.d/`.

phated:
@ngdangtu-vn Does this change the behavior people expect from the import syntax in node? I don't really want to define our own behavior for that.

ngdangtu-vn:
> Does this change the behavior people expect from the import syntax in node?
Yes and no:

- Yes because the import of NodeJs couldn't read dir in `dir/index.mjs` structure.
- No because this only affect on how CLI load files, and CLI only. It means you what happens inside gulpfile still follows every ESM rules in NodeJs. Moreover, from the first place, I don't think people will put much thought about how CLI load files.

Therefore, this patch is still safely to merge.

phated:
@ngdangtu-vn I believe this needs conflicts resolved since I merged #239

ngdangtu-vn:
@phated It has been too long so I don't really remember what I did back then :)) I can resolve this but do we really need this PR anymore?

phated:
That's fine. I wasn't sure I wanted to include this so we can just close it out.

PR #233, opened 2021-08-28
Update yargs version to 15.4.0

hmartinez69:
This PR goal to fix vulnerabity security.

yargs-parser is a dependency of yargs which have a "Prototype Pollution" fixed on >=18.1.2 versions. yargs call this new version since its 15.4.0 release.


phated:
https://overreacted.io/npm-audit-broken-by-design/

PR #252, opened 2024-01-21
fix: remove moveDrive logic in test

### LINKED ISSUE #247: Check `moveDrive` logic on Windows with use of exec `cwd`
> Do we still need this `moveDrive` logic if we are using `cwd`? Not going to block on it

_Originally posted by @phated in https://github.com/gulpjs/gulp-cli/pull/239#discussion_r1444067723_

sttk:
Resolves #247.

This moveDrive logic was implemented to avoid an error in a test case. Now, that test case passes without this logic.

### DIFF
--- a/test/execution-errors.js
+++ b/test/execution-errors.js
@@ -43,13 +43,8 @@ describe('execution error', function() {
   });
 
   it('should output an error if gulp is not found', function(done) {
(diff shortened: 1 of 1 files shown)


phated (approved):
Thanks!

PR #254, opened 2024-02-14
chore: Add deprecated warning for gulplog v1 messages

### LINKED ISSUE #249: Consider wiring gulplog v1 handlers and print deprecation warning when used.
gulplog v2 doesn't see gulplog v1 messages (see https://github.com/gulpjs/gulplog/issues/17). We should either implement a workaround here (or in gulplog itself) where the old messages are handled but print a deprecation warning.

phated:
Closes #249 

This wires up gulplog v1 handlers and prints a deprecation message when they are used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "copy-props": "^4.0.0",
     "fancy-log": "^2.0.0",
     "gulplog": "^2.0.1",
+    "gulplog-v1": "npm:gulplog@1.0.0",
     "interpret": "^3.1.1",
     "liftoff": "^4.0.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -34,6 +34,7 @@
     "copy-props": "^4.0.0",
     "fancy-log": "^2.0.0",
     "gulplog": "^2.0.1",
+    "gulplog-v1": "npm:gulplog@1.0.0",
```
phated, line 37: I'm going to create a scoped package for this instead of using aliases.

phated:
Replaced by #266

PR #259, opened 2024-03-17
feat!: Upgrade to Liftoff v5 and avoid merging flags/config/env

phated:
This upgrades to Liftoff v5 and removes a bunch of the merging between flags, config, and the environment.

I've also decided that I don't want to merge the local config and home config, so this changes to find the first available config.

Notice that Liftoff v5 ensures that the `cwd` is the directory where the gulpfile as specified in the `config` exists. Since  `gulp.src`/`gulp.dest` are often used with relative paths, it is important that the `cwd` match the location of the gulpfile. This is what affected the majority of the test changes.

TODO:
* [x] Doc updates


phated:
@sttk please review this. I've cherry-picked it into my translations prototype to ensure these changes work with the concept.

sttk (commented):
@phated I've reviewed it, and it looks good! Great!

PR #260, opened 2024-03-17
feat: Support theming and translations via config files

### LINKED ISSUE #63: Allow customizing CLI colors through .gulprc file

### LINKED ISSUE #177: Log only the "Finished" messages and not the "Starting" ones
Hey, it would be really useful if I could tell Gulp to log only when a task is done, not when it starts. 
I don't actually care when the task starts during development. I only care about how much time it took and if it ran successfully.

I've tried the `log-level` option and it just turns the starting/finished logs on or off.

What do you think?

### LINKED ISSUE #206: Remove leading ellipsis in .flags string when running --tasks
Hello,
Not a bug but a question and suggestion here. I hope i'm in the right repository.

Why there is leading ellipsis added on the `task.flags` description when running `gulp --tasks`? Can we just remove it? So we can add whatever if we want or keep it simply clean.

image

Thanks!

phated:
This implements theming and translation support via the `.gulp.js` config files. If a user supplies a `message` function in their config file, we will call it with various message symbols and extra data. The user can return the message they want to print, return `false` explicitly to silence the message, or do nothing to fallback to the gulp-cli default message.

Since anything logged via `gulplog` will pass through this mechanism, users can also intercept string messages from plugins and return something different.

Users can also override the timestamp we print by providing a `timestamp` function in their config file.

Closes #63
Closes #177
Closes #206
Closes #161 (replaces it)


lib/versioned/^4.0.0/log/check-task-not-found.js
phated, line 1: This idea is amazing and I want to cherry-pick the change into a separate commit. I want to attach properties to the error instead of building a new object. I'm also going to add those properties to the new undertaker before I release it.

index.js
```diff
@@ -49,38 +52,34 @@ var cli = new Liftoff({
...
+  .fail(onFail)
   .options(cliOptions);
 
 var opts = parser.parse();
 
 // Set up event listeners for logging temporarily.
-toConsole(log, opts);
+// TODO: Rework console logging before we can set up proper config
+// Possibly by batching messages in gulplog until listeners are attached
+var cleanupListeners = toConsole(log, opts, buildTranslations());
```
phated, line 69: This is a general bug with our libraries. I'll create a "known issue" on gulp-cli for us to fix this in our next breaking version.

phated:
Responding to https://github.com/sttk/gulp-cli/pull/3#issuecomment-1987137214

> @phated Your idea is very interesting, but there are some concerns.
> 
> 1. There are cases that having to provide a plugin to change messages or styles would be burdensome. Some users might want to change only colors of certain words in certain messages against the background color of their terminal. (In fact, I consider that this case is the most important request regarding this theming feature.) It should be possible to change colors with minimal effort. As an addition to your idea, it can be considered to enable to override the `getMessage` function in a config file, but the config file would be limited to `.js` file only.

I'm okay making the theming and translations to be a little more complex for users if it keeps our implementation straightforward and easy to modify. Users can provide the `message()` function and only handle a single message (like `messages.DESCRIPTION`, handle every symbol, or any amount of message in between. This implementation also makes it easy to silence specific messages.

I've change the `.gulp` config file lookup to only support `interpret.jsVariants` in #261 so users always have access to this feature.

> 2. `fancy-log` always outputs the timestamp at the beginning of the line, and its color or format cannot be changed. There are indeed requests in the issues for changing the color against the terminal background color, so we need to provide the solution for this.

I've removed `fancy-log` and reimplemented the `Timestamp` class that I used there. Users can provide a `timestamp()` function in the config file to override the timestamp.

We will probably deprecate `fancy-log` post-v5 if we're not using it anymore.

> 3. There are cases that CLI options `—color` and `—no-color` would not be supported if the external plugin uses a coloring module other than `chalk`.

Users could intercept this in their `message()` function and recreate the messages if it bothers them.

> 4. In the draft version, the help messages and task tree are not configurable. For these as well, I believe it's necessary to make it possible to change the colors at least against the terminal's background color.

These are completed now. What do you think?

lib/shared/log/tasks.js
```diff
@@ -71,55 +77,65 @@ function logTasks(tree, opts, getTask) {
...
-      lines.push(line);
-      line.label = chalk.white(flagBars) + chalk.magenta(ent[0]);
+      line.label = flagBars + ent[0];
+      line.width = stringWidth(line.label);
 
-      maxLabelWidth = Math.max(maxLabelWidth, line.label.length);
+      maxLabelWidth = Math.max(maxLabelWidth, line.width);
 
-      if (typeof ent[1] !== 'string' || !ent[1]) return;
-      line.desc = chalk.white('…' + ent[1]);
```
phated, line 122: Just removing the `...` and coloring here. People can create whatever theming they want inside their gulpfiles.

lib/shared/log/tasks.js
```diff
@@ -71,55 +77,65 @@ function logTasks(tree, opts, getTask) {
...
     } else {
-      taskBars += '┬ ';
+      taskBars += translate.message(messages.BOX_DRAWINGS_LIGHT_DOWN_AND_HORIZONTAL);
     }
+    taskBars += ' ';
 
     var line = {};
     if (task.depth === 1) {
-      line.label = chalk.white(taskBars) + chalk.white(task.label);
+      line.label = taskBars + task.label
```
phated, line 93: Not allowing people to theme the task names since those should be verbatim for the gulpfile. They can name tasks them in a different language already and I've removed the coloring all together.

lib/shared/log/tasks.js
```diff
@@ -71,55 +77,65 @@ function logTasks(tree, opts, getTask) {
...
     }
+    flagBars += ' ';
 
     Object.entries(task.flags).sort(flagSorter).forEach(addFlagsToLines);
 
     function addFlagsToLines(ent) {
       if (typeof ent[0] !== 'string' || !ent[0]) return;
       var line = {};
-      lines.push(line);
-      line.label = chalk.white(flagBars) + chalk.magenta(ent[0]);
```
phated, line 117: Same as above. No longer allowing theming here.

lib/shared/translate.js
```diff
@@ -0,0 +1,267 @@
...
+      if (message === false) {
+        return '';
+      }
+
+      // If the user hasn't returned a message or silenced it with `false`
+      // get the default message. Will return the first argument if the message
+      // is not defined in the `@gulpjs/messages` package
+      return getDefaultMessage(msg, data);
+    },
+    timestamp: function () {
```
phated, line 246: I considered passing the message symbol here, but I think that we just need to separate timestamp vs non-timestamp output by the usage of `log` vs `console.log`

lib/versioned/^4.0.0/index.js
```diff
@@ -77,16 +71,20 @@ function execute(env, cfg, opts) {
...
       var runMethod = opts.series ? 'series' : 'parallel';
       gulpInst[runMethod](toRun)(function(err) {
         if (err) {
           exit(1);
         }
       });
     } catch (err) {
-      log.error(chalk.red(err.message));
-      log.error('To list available tasks, try running: gulp --tasks');
+      var task = checkTaskNotFound(err);
```
phated, line 82: @sttk do you want to pull this into a separate PR?

messages.js
phated, line 1: This will become the new module so users can install it for usage in their `.gulp.js` files.

test/fixtures/config/theming/MISSING_GULP/.gulp.js
```diff
@@ -0,0 +1,16 @@
...
+    }
+
+    if (msg === Symbol.for('GULP_CLI_NPM_INSTALL_GULP')) {
+      // Silence for test
+      return false;
+    }
+  },
+  timestamp: function () {
+    return false;
+  }
```
phated, line 15: @sttk should we leverage this feature in all of the tests to avoid the need for the `eraseTime` utility?

phated:
@sttk this is ready for review. Please take a look! I've updated the top comment to describe the feature and updated tests, etc. Thank you for all of the amazing work that I used for the basis of these changes 🙇 ❤️ 

When we finalize this, I will put together the `@gulpjs/messages` package and update all of the TODOs.

test/config-message-function.js
```diff
@@ -0,0 +1,462 @@
...
+
+    function cb(err, stdout, stderr) {
+      expect(err).toBeNull();
+      expect(stderr).toEqual('');
+      expect(stdout).toEqual(expected);
+      done(err);
+    }
+  });
+
+  // TODO: Do we want tests returning false for every message?
```
phated, line 67: @sttk what do you think about this?

lib/versioned/^4.0.0/index.js
sttk: Yes. Ideally, it would be better to modify `undertaker` to set these task names as properties of the error to avoid reverse parsing from the error message. Can I modify `undertaker` for this purpose?

lib/shared/translate.js
sttk: I think it's better that `message` function always returns a line and receives message parts via `data` parameter, and timestamp is passed by including `data` parameter, too.
By doing so, we and users can choose whether attaching timestamp before a message line in the`message` function.

lib/shared/log/tasks.js
sttk: This depends on https://github.com/gulpjs/gulp-cli/pull/260#discussion_r1528380052

lib/shared/log/tasks.js
sttk: This depends on https://github.com/gulpjs/gulp-cli/pull/260#discussion_r1528380052

test/config-message-function.js
sttk: If `message` function always a line, and outputting the line is controlled in common process, the test is sufficient for just one message, I think.

sttk:
@phated I have replied to your comment. Have a look them.

lib/versioned/^4.0.0/index.js
phated: We will need this transformation for gulp v4 even after we implement the change in undertaker for gulp v5

lib/shared/translate.js
phated: I don't want to pass message parts but I'm considering attaching the timestamp in the `message` function. Please let me consider which is preferable.

lib/versioned/^4.0.0/index.js
sttk: I understood. I'll create a new PR of this `checkTaskNotFound` function for gulp v4.

index.js
phated: This could be a temporary solution, but I want people to be able to use typescript, babel, and other loaders so I think we'll just create the known issue.

I'm already planning the changes to Liftoff to support this, but we need top-level await (in gulp 6).

lib/shared/translate.js
phated: I've been considering your proposal, but I was struggling with some ideas.

I want users to be able to easily disable timestamps if they don't like them. If we attach timestamps in the `message` function, users would have to override all messages to disable timestamps.

Instead, I've decided to pass the message into the `timestamp` function so users can vary the timestamp based on the message. However, people are able to ignore the parameters and return a single timestamp.

lib/versioned/^4.0.0/index.js
phated: Thank you. When will you have time to submit this?

lib/shared/log/tasks.js
phated: Can you explain?

lib/shared/log/tasks.js
phated: Can you explain?

phated:
@sttk please review again now that I've addressed all your comments.

lib/shared/log/tasks.js
phated: I'm starting to think these need to pass through `translate.message` because it's annoying for users to theme the keys on `task.flags = {}`

lib/versioned/^4.0.0/log/check-task-not-found.js
phated, line 1: Related undertaker PR: https://github.com/gulpjs/undertaker/pull/102

lib/versioned/^4.0.0/index.js
sttk: I've create the PR for this. → #262

phated: Thank you! I've merged it into this branch

lib/shared/log/tasks.js
phated: @sttk I've added more messages so I can translate this.

lib/shared/log/tasks.js
phated: @sttk I've added more messages so I can translate this.

phated:
@sttk I've address the comments and added documentation. Can you provide a final review so I can make the `@gulpjs/messages` package and finish this PR in the next 3 days? 🙏

sttk:
@phated I have added [this comment](https://github.com/gulpjs/gulp-cli/pull/260#discussion_r1535573944), so please take a look. Otherwise, I think everything is good.

lib/shared/translate.js
phated: There is no way for the user to access getDefaulMessage. If they return nothing, the default message is used. If they return `false`, the message is silenced.

sttk: Sorry. What I wanted to do in the second example above can be written as follows.
```
    getMessage = function(data) {
        switch (data.tag) {
        case messages.PRELOAD_BEFORE:
            return false;   // message isn't displayed.
        case messages.MISSING_GULPFILE:
            return timestamp() + chalk.yellow('Gulpfile was not found!')  + ' 😱😱😱' ;
        default:
            return ''; // the default message is displayed.
        }
    }
```

sttk: What I really wanted to suggest is passing the `timestamp` function to the `data`, and then in the message function, appending it (or removing it if the user doesn't want it).

phated: I don't want to attach timestamp to the data object. I don't want to require users to implement the `message()` function if they only want to remove the timestamp.

phated: To be clear, I want it to be 1-3 lines of JS to disable timestamps. With this implementation that is possible.

sttk: Understood, I think that's fine.

phated:
@sttk Thank you for the thorough review! I am going to merge this now.

test/fixtures/config/theming/MISSING_GULP/.gulp.js
phated: #263

index.js
phated: #264

PR #262, opened 2024-03-21
update: separate error while running gulpfile into `task not found` and `fail to run`

sttk:
This PR separates the error while running gulpfile into `task not found` and `fail to run`, and adds a test case for the ‘task not found’ error. 
However, the test case for `fail to run` error is not added because I could not find the case to occur `fail to run` error.


phated (approved):
Thanks @sttk! I've updated the PR to normalize the error properties like in Undertaker. This looks good to go 👍

PR #265, opened 2024-03-23
feat: Add versioned handler for gulp v5

### LINKED ISSUE #250: Add CLI handler for gulp v5
We'll need to include the gulp 5.0.0 semver range in the list of runners; otherwise the gulp v5 release won't be handled by the CLI.

phated:
This is mostly a copy-paste of the `^v4.0.0` directory, but I removed the `normalizeError` helper since we made the change in undertaker.

@sttk Please review and let me know if anything else needed to change for gulp v5.

Closes #250

### DIFF
--- a/lib/versioned/^5.0.0/format-error.js
+++ b/lib/versioned/^5.0.0/format-error.js
@@ -1,0 +1,24 @@
+'use strict';
+
+// Format orchestrator errors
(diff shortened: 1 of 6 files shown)


sttk (approved):
Good work! I think it's fine.

PR #266, opened 2024-03-23
feat: Add deprecated warning for gulplog v1 messages

### LINKED ISSUE #249: Consider wiring gulplog v1 handlers and print deprecation warning when used.

phated:
This is an alternative for #254 that relies on changes made in [sparkles](https://github.com/gulpjs/sparkles/pull/10), [glogg](https://github.com/gulpjs/glogg/pull/9), and [gulplog](https://github.com/gulpjs/gulplog/commit/4999733621311f409bf29673361315ee023cb261) to listen & re-emit on the new global namespace.

Closes #249

TODO:
- [x] Add the message symbol to `@gulpjs/messages` package


package.json
```diff
@@ -47,6 +47,7 @@
     "@babel/core": "^7.20.2",
     "@babel/preset-env": "^7.20.2",
     "@babel/register": "^7.18.9",
+    "@gulpjs/gulplog-v1": "npm:gulplog@1.0.0",
```
phated, line 50: I'm fine with this because I'm mapping to the `@gulpjs` scope and this is only a devDep for the tests

lib/shared/log/to-console.js
```diff
@@ -37,12 +40,33 @@ function toConsole(log, opts, translate) {
...
+    log.removeListener('deprecated', onDeprecated);
     log.removeListener('error', onError);
     log.removeListener('warn', onWarn);
     log.removeListener('info', onInfo);
     log.removeListener('debug', onDebug);
   };
 
+  function onDeprecated() {
+    if (!deprecatedPrinted) {
+      var msg = { tag: Symbol.for('GULP_CLI_GULPLOG_DEPRECATED') };
```
phated, line 52: By allowing this to pass through the translation mechanism, users that don't want to see it can return `false` to silence the warning :tada:

test/config-message-function.js
```diff
@@ -472,4 +472,19 @@ describe('config: message function', function() {
...
+    var cwd = path.join(baseDir, 'GULPLOG_DEPRECATED');
+    var expected = 'GULPLOG V1 IS DEPRECATED\n';
+
+    var opts = { cwd: cwd };
+    exec(gulp(), opts, cb);
+
+    function cb(err, stdout, stderr) {
+      expect(err).toBeNull();
+      expect(stdout).toEqual('');
+      expect(stderr).toEqual(expected);
```
sttk, line 486: Shouldn't the message the plugin attempted be output with deprecated message?

phated:
@sttk please review right away 🙏 This will be the last PR into gulp-cli before I release v3.0.0 and prepare gulp for the v5 release. Thank you for dedicating time!

sttk (commented):
@phated I've reviewed. Please take a look as there is one I pointed out.

test/config-message-function.js
phated: I silenced all other messages via the config file. Do you think I need to allow the other message?

sttk: I believe `gulplog-v1` is intended for messages output by existing plugins. Wouldn't it be problematic if those messages are not output?

Why isn't the original message (`"Some error here"` in this test case) being output after the deprecated message in [this part of the code](https://github.com/gulpjs/glogg/blob/master/index.js#L21-L23)?

phated: I think you don't understand how the translation code works. If `false` is returned, nothing is printed.

sttk: I understood why the original error message is not being output.

If the message function returns false, is it possible all messages from existing plugins not to be output, too? Or, are there plans to control the output of plugin messages using this message function in the future?

phated: Any messages to gulplog can be controlled through the `message` function, as you've seen here. Generally plugin errors will just be strings, not symbols, so users will have to match string values.

sttk: I see. However, in actual cases, most of times plugin messages will be output as they are. Therefore, I think it would be better to add test cases where the original message is output.

phated: This test is a unit test to ensure the deprecate message can be translated. There may be a case for creating another test to test the original message but I don't want to dig into where that can fit in the harness. This is a good enough test for now.

PR #271, opened 2025-03-22
fix: update DESCRIPTION message according to the description option

folliehiyuki:
The `description` option in `.gulp.*` configuration file is never read.


phated:
It looks like I never removed the docs for the `description` property when converting everything to the translation code. I think we'll probably just remove the docs since we want all of the message logic to go through the translation mechanism.

Do you want to update this PR to remove those docs?

folliehiyuki:
Is it just in the README? Some test files still use `exports.description` (https://github.com/gulpjs/gulp-cli/blob/master/test/fixtures/config/foo/bar/quux/.gulp.babel.js, https://github.com/gulpjs/gulp-cli/blob/master/test/fixtures/config/foo/bar/.gulp.js, https://github.com/gulpjs/gulp-cli/blob/master/test/fixtures/config/qux/.gulp.js) and they don't seem to be referenced anywhere. Do they need to be removed as well?

phated:
I think we need to remove both the readme and unused test files. I forgot to remove both with my translation work.

folliehiyuki:
I deleted all the relevant files and `npm run test` still passed on my local machine.

PR #272, opened 2025-06-01
chore: Fix CI

phated:
Checking to see what CI stuff is broken.


phated:
The Node 24 failures will be resolved by https://github.com/gulpjs/liftoff/pull/134 which I'll get merged and released.

PR #275, opened 2026-02-22
Fix: upgrade rimraf (from deprecated v3.0.2 to v5.0.7) 

EvilSpirit39:
rimraf v3.0.2 is deprecated, so upgraded to supported version 5.0.7.
For upgrading, rimraf.sync is no longer available and switched to rimrafSync.

image


EvilSpirit39:
CI failed. I'll check the reason

EvilSpirit39:
The reason is Node.js version.
new version of rimraf supports after Node.js v14.18. 
v12 or prior are not supported.

https://github.com/isaacs/rimraf/blob/fb974adeed36bfc4d41a601f60a751aa509dbe2a/package.json#L73

